PR #5, opened 2017-12-29
resolves issue 3

boothby:


boothby:
got stale, new request forthcoming

PR #44, opened 2018-06-14
Put Python examples in intro

JoelPasvolsky:


bellert:
Should the examples be removed from the README so they aren't duplicated in the documentation?

Issue #50, opened 2018-07-09
Fails to build when Boost / BoostGraph is present


amccaskey:
See attached log file. 

Basically there is ambiguity between std::begin/std::end when integrating minorminer into another library where Boost/BoostGraph is present. Also, there is a clash between this libraries graph namespace and that from Boost. 

[make.txt](https://github.com/dwavesystems/minorminer/files/2177133/make.txt)

I have a fix in for this at https://github.com/amccaskey/minorminer. I will issue a pull request.

boothby:
I've merged your pull request, thanks very much!

PR #51, opened 2018-07-09
Enabling integration with libraries that leverage Boost and Boost Graph

amccaskey:
This PR addresses Issue #50


boothby:
Thanks for your contribution!  As I paw through the diffs, I notice that that there's still, e.g., `using std::set`.  Should that go, as well?

amccaskey:
Whoops, missed the set -> std::set. It builds without it, but its good practice to preface the STL containers with the std namespace. I just pushed an update.

PR #57, opened 2018-08-31
chainlength updates & miner class

boothby:
Sorry for the massive PR...
* restructuring of find_embedding and pathfinder, to support interface wrappers
* new feature, max_beta, which can be used to tune chainlength and success probability (in general, it seems that the lowest max_beta to yield successes will give the shortest chains)
* other improvements to chainlength performance
* added miner class, with features geared at "mining" for minors... maybe the name isn't just a bad pun now?


boothby:
stale

PR #61, opened 2018-10-10
release candidate

boothby:
I think this version is pretty solid, and hopefully I've fixed all the CI issues...


include/embedding.hpp
```diff
@@ -113,7 +117,9 @@ class embedding {
     inline int weight(int q) const { return qub_weight[q]; }
 
     //! Get the maximum of all qubit weights
-    inline int max_weight() const { return *std::max_element(std::begin(qub_weight), std::begin(qub_weight) + num_qubits); }
+    inline int max_weight() const {
+        return *std::max_element(std::begin(qub_weight), std::begin(qub_weight) + num_qubits);
```
randomir, line 121: why not `std::end`?

include/embedding_problem.hpp
```diff
@@ -196,24 +251,87 @@ class embedding_problem_base {
...
+        initialized = embedded = desperate = target_chainsize = improved = 0;
+    }
+
+  private:
+    //! computes an upper bound on the distances computed during tearout & replace
+    int compute_margin() {
+        auto max_degree =
+                (*std::max_element(begin(var_nbrs), end(var_nbrs), [](const vector<int> &a, const vector<int> &b) {
+                    return a.size() < b.size();
+                })).size();
```
randomir: maybe `->size()` is more readable?

include/embedding_problem.hpp
```diff
@@ -196,24 +251,87 @@ class embedding_problem_base {
...
+        shuffle(std::begin(var_nbrs[u]), std::end(var_nbrs[u]));
+        return var_nbrs[u];
+    }
+
+    //! a vector of neighbors for the variable `u`, applying a random
+    //! transposition before returning the reference
+    const vector<int> &var_neighbors(int u, rndswap_first) {
+        if (var_nbrs[u].size() > 2) {
+            int i = randint(var_nbrs[u].size() - 1);
+            std::swap(var_nbrs[u][i], var_nbrs[u][i + 1]);
```
randomir, line 328: off by one?

randomir: Ok, it's not. But it's confusing because internal `randint` behaves differently from [`experimental::randint`](https://en.cppreference.com/w/cpp/experimental/randint).

include/embedding_problem.hpp
```diff
@@ -229,36 +347,6 @@ class embedding_problem_base {
...
-    }
-
-    //! print at the debug verbosity level (only works when `CPPDEBUG` is set)
-    template <typename... Args>
-    void debug(const char *ONDEBUG(format), Args... ONDEBUG(args)) const {
-        ONDEBUG(params.debug(format, args...));
-    }
-
     //! make a random integer between 0 and `m-1`
     int randint(int m) { return rand(params.rng, typename decltype(rand)::param_type(0, m - 1)); }
```
randomir: Maybe change to `randint(int a, int b)` to make it clearer?

include/find_embedding.hpp
```diff
@@ -12,134 +12,190 @@
...
-        int all_chains = 1;
-        for (auto &vC : params_.initial_chains) {
-            // don't bother putting in initial chains that will be overwritten by fixed chains
-            if (!var_fixed_unscrewed[vC.first]) {
-                std::vector<int> C;
-                all_chains &= qubit_components.into_component(c, vC.second, C);
-                params.initial_chains.emplace(screw_vars[vC.first], C);
+    vector<int> _filter(int n, vector<int> &test) {
+        vector<int> r(n);
+        for (int i = 0, front = 0, back = num_vars - num_reserved; i < n; i++) {
```
randomir: since `num_vars` and `num_reserved` are public (easy to modify from outside), maybe a range check here? or below under if/else?

include/find_embedding.hpp
```diff
@@ -12,134 +12,190 @@
...
+            if (test[i]) {
+                r[back++] = i;
+            } else {
+                r[front++] = i;
```
randomir: since `back`/`front` might be out of bounds, better use `.at()` here

include/find_embedding.hpp
```diff
@@ -12,134 +12,190 @@
...
-            all_chains &= qubit_components.into_component(c, vC.second, C);
-            params.restrict_chains.emplace(screw_vars[vC.first], C);
+        return r;
+    }
+
+    vector<int> _inverse(vector<int> &f) {
+        int n = f.size();
+        vector<int> r(n);
+        for (int i = n; i--;) {
+            r[f[i]] = i;
```
randomir: safer with `r.at(f[i]) = i`

include/pathfinder.hpp
```diff
@@ -232,6 +261,10 @@ class pathfinder_base {
     //! lower the maximum chainlength
     int improve_chainlength_pass(embedding_t &emb) {
         bool improved = false;
+        dijkstras[0].reorder(params.rng);
+        for (int v = num_vars + num_fixed; v-- > 1;) {
```
randomir: isn't this equivalent to the (standard): `for (int v = ...; v > 1; v--)`?

randomir: also, is size of `dijkstras` >= `num_vars + num_fixed`? if `num_vars + num_fixed == 1`, should you skip the loop?

matlab/make.m
```diff
@@ -1 +1 @@
-mex find_embedding.cpp -I../include CXXFLAGS="-std=c++11 -fPIC -O0" -lut -largeArrayDims -v
+mex find_embedding.cpp -I../include CXXFLAGS="-std=c++1y -fPIC -O3" -lut -largeArrayDims -v
```
randomir, line 1: It's probably better to use `c++14` now (after the standard became official).

randomir (commented):
This is a huge PR, practically amounts to a rewrite. Without going into `minorminer` details (which would probably take weeks), in a couple of hours, I've been only able to take a cursory look...

include/embedding.hpp
boothby: because it's got length num_qubits+num_reserved

matlab/make.m
boothby: My understanding is that Adam tried that, and it broke several compilers that we support.  I've deliberately avoided use of c++14 features for that reason.

include/pathfinder.hpp
boothby: 1) yes, it looks like this works just as well?
2) i'm not going to waste a conditional at this point in the code to check if the overall problem is utterly trivial.  otoh... your suggested form for the loop apparently cuts out a postdecrement in the final iteration (I say "apparently" because modern compilers are too smart for that)

matlab/make.m
randomir: I don't know which compilers we're supporting, but there's [this](https://stackoverflow.com/questions/28825113/difference-between-gcc-compile-options-std-c1y-and-std-c14).

include/pathfinder.hpp
randomir: ad 2) no, i wasn't suggesting you test for the trivial case; but depending on the answer to the first question (what's the size of `dijkstras` array), suggesting an error if you need to update `dijkstras[num_vars + num_fixed]`.

include/embedding.hpp
randomir: :+1:

matlab/make.m
boothby: Hmm.  Now I really don't know.  My suspicion is that Visual Studio was the source of problems and I might have broken it with the change to +1y

include/pathfinder.hpp
boothby: `dijkstras` has size `num_vars+num_fixed`; it'd be an error to access that location ;)

matlab/make.m
boothby: in my personal opinion, both windows and matlab should be burned and stricken from history books but since I can't control what platforms our users are on...

include/pathfinder.hpp
randomir: then ignore my second comment...

matlab/make.m
randomir: According to [this](https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=vs-2017), `/std` option was introduced only in MSVC 2017, and it supports c++14. I'm not sure if and/or how `-std=c++1y` works on M$ at all. Not my platform of choice either. :)

include/embedding_problem.hpp
boothby: yeah that seems prudent

include/embedding_problem.hpp
boothby: yup

include/embedding_problem.hpp
boothby: yeah that seems prudent

include/find_embedding.hpp
boothby: i've changed this function a little bit; please review thew new version when i push my patch

include/find_embedding.hpp
boothby: I disagree (see new version)

include/find_embedding.hpp
boothby: true true

include/embedding_problem.hpp
boothby: yeah; fixed that

matlab/make.m
boothby: I'm going to see if I can't test `std=c++14` on these platforms... I know Adam had some environments set up for that but this might take some digging

randomir: :+1:

randomir (approved):
:+1:

arcondello (approved):

boothby:
Aidan gave a verbal OK on the algorithmic improvements; closing so we can move on.  I'd like to see some more testing on exotic platforms... but that's looking like a bit of a project.  Hopefully we get a bug report if something's broken.

PR #67, opened 2018-12-12
Fixes issue 64 & 66, bumps version #

boothby:


include/embedding_problem.hpp
```diff
@@ -285,7 +285,7 @@ class embedding_problem_base {
                 std::max_element(begin(var_nbrs), end(var_nbrs), [](const vector<int> &a, const vector<int> &b) {
                     return a.size() < b.size();
                 })->size();
-        return max_degree * num_q;
+        return max(max_degree, 1UL) * num_q;
```
randomir: Looks like this doesn't work in older compilers. Maybe like this?
```
return std::max(max_degree, 1) * num_q;
```
or:
```
return std::max<unsigned long>(max_degree, 1UL) * num_q;
```

PR #74, opened 2019-01-15
Performance update

boothby:
This update results from an observation that in node-weighted Dijkstra, the first time a node is discovered, it's always a shortest path.  This update is fairly complex, because we no longer require a decrease-key operation -- the principal data structure, a pairing heap, can be simplified to a 2-pointer structure (instead of the 3 required for a fast decrease-key).  We take the opportunity to improve unit testing, and add randomized tests.  This update is impactful because these queue operations represent about half of the runtime.


boothby:
smoketest failures merit further investigation

PR #77, opened 2019-02-11
Performance update

boothby:
* Profiled and hacked down the hotspots by refactoring pairing queue datastructure
* Fully parallel implementation of find_chain -- still need to parallelize find_short_chain


boothby:
Giving up on find_short_chain for the time being.  Unlike find_chain, that algorithm isn't trivially parallelizable.

PR #78, opened 2019-03-02
EmbeddingInfo container class for recording post-hoc info.

jberwald:
Idea is to use this later to also return chain connection info, etc., to the user. 

While the `EmbeddingInfo` object is fairly minor, I hope it can get the ball rolling on adding chain information to an object returned by minorminer.  

@boothby

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -174,5 +174,7 @@ venv.bak/
 # mypy
 .mypy_cache/
 
+# PyCharm
+.idea/
 
 # End of https://www.gitignore.io/api/c,c++,python
--- a/python/minorminer.pyx
+++ b/python/minorminer.pyx
@@ -41,6 +41,8 @@ This implementation adds several useful features:
 """
 include "minorminer.pxi"
 import os
+from collections import Counter
+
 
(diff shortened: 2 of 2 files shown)


boothby:
I'm not against adding embedding metrics to minorminer, but this breaks the interface (as noted by circleci).  The downstream impact of this change would be huge.  If you look in the miner class, you'll see that I've put some thought into comparisons of embeddings.  If you're keen to make this happen, I'd prefer to keep the richer embeddings optional -- something like
```
from minorminer import find_embedding, RichEmbedding
...
emb = RichEmbedding(find_embedding(S, T, **options))
```
where the `RichEmbedding` class should implement a `dict` interface (except it's fine to be immutable; maintaining derived statistics can be a pain)

jberwald:
I'm happy with the external wrapper class method. It's purely for convenience. The other option is to add a flag to either return a `RichEmbedding` object or not. I think I lean towards your method, though.

boothby:
Stale; it doesn't look like this is much of a value-add.  If anything, this should probably be a feature for the `minorminer.utils` package.

PR #81, opened 2019-05-01
Examples of suspend_chains

joseppinilla:
Just thought it would be useful to have.
But also, wanted to point out that, since the internal pins are also returned in the embedding, a naive use of the embedding with the drawing capabilities, will cause an error if show_labels=True is used.

Maybe there's nothing to do here but the user should be aware of this. Or maybe the drawing package should handle it... or... networkx shouldn't use label keys to address the pos dictionary, which is what causes the error.

### DIFF
--- a/examples/suspend_chains.py
+++ b/examples/suspend_chains.py
@@ -25,7 +25,4 @@ blob_C0 = [6,7,14,15]
 suspend_chains = {'A': [blob_A0, blob_A1],'C':[blob_C0]}
 embedding = minorminer.find_embedding(K3, C, suspend_chains=suspend_chains)
 plt.subplot(2,2,4)
-# ISSUE: show labels in networkx iterates through labels and uses key in
-# pos. If label is given but not pos, there's an error.
-# This line fails.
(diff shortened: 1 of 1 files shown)


arcondello (approved):

arcondello (changes requested):
I like the example, but we should add a test to CI so that we know it doesn't get out of date.

boothby:
> But also, wanted to point out that, since the internal pins are also returned in the embedding

Uhhh... can you rephrase that in the form of a bug report?  Because that's awful and we shouldn't be returning those.  I'll have time to fix that late next week

boothby:
I've added these examples as tests in response to @arcondello's change request, and patched the bug in PR #88 .

boothby:
I've submitted a pull request to @joseppinilla's repo removing the note about the bug resolved in #88.  @arcondello's changes are resolved and I'm happy with this example (provided that note is removed).

boothby (changes requested):
Strike lines 28-30 (accepting my PR to @joseppinilla's repo)

boothby (approved):

arcondello (approved):

PR #82, opened 2019-05-02
Visible parameters of the embedding method

joseppinilla:
Is this something that could be considered?
Looking ahead into having other embedding methods, or more "sophisticated" `Composites `that can make use of the flexibility of `minorminer`, it would be very useful to be able to read out the parameters and eventually add them to the sampler parameters. e.g:
```
class ???EmbeddingComposite(...
...
param = self.child.parameters.copy()
param['embedding_parameters'] = self._embedding_method.parameters.copy()
...
```

```
# Example usage
import minorminer

print(minorminer.parameters)
>>> {'chainlength_patience',
 'fixed_chains',
 'initial_chains',
 'inner_rounds',
 'max_beta',
 'max_fill',
 'max_no_improvement',
 'random_seed',
 'restrict_chains',
 'return_overlap',
 'skip_initialization',
 'suspend_chains',
 'threads',
 'timeout',
 'tries',
 'verbose'}

```

### DIFF
--- a/python/minorminer.pyx
+++ b/python/minorminer.pyx
@@ -42,6 +42,11 @@ This implementation adds several useful features:
 include "minorminer.pxi"
 import os
 
+parameters = {"max_no_improvement", "random_seed", "timeout", "tries", "verbose",
+         "fixed_chains", "initial_chains", "max_fill", "chainlength_patience",
+         "return_overlap", "skip_initialization", "inner_rounds", "threads",
+         "restrict_chains", "suspend_chains", "max_beta"}
+
 def find_embedding(S, T, **params):
     """
     find_embedding(S, T, **params)
@@ -588,4 +593,4 @@ cdef _read_graph(input_graph &g, E):
         g.push_back(L[a],L[b])
     return L
 
-__all__ = ["find_embedding", "VARORDER", "miner"]
+__all__ = ["find_embedding", "VARORDER", "miner", "parameters"]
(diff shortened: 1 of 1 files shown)


arcondello:
I am wondering if we should just be relying on Python's [inspect](https://docs.python.org/3/library/inspect.html) module for this sort of thing. I am worried about the maintenance. @boothby what do you think?

Another approach might be to decorate the different embedding functions that share features. So

```
@acceptstimeout
def find_embedding1(..., timeout=100):
   ...
```

joseppinilla:
I agree, the way I did it would be hard to maintain. It's mostly an example of an implementation and how to use it. Unless it only has one dictionary, and not two, i.e. `names` and `parameters`.

Would `inspect `work? I think Cython isn't supported in inspect, and also, the parameters are only parsed from `**params` not as implicit function arguments.

boothby:
Sorry for the delay, I've been bogged down with other work... I don't mind this change at all.  When I add a parameter, it requires changing about a dozen lines of code across something like 5 files.  So it's already a nightmare; this PR only moves a line and my workflow already begins with `grep`.  My temptation is to pitch the review back to @arcondello -- how do you feel about exposing parameters to EmbeddingComposite in this way?

boothby:
Okay this issue came to a head today... again I apologize for the extended delay.  Alex and I had a chat and I decided that this was a perfect grenade-jumping exercise.  Cython doesn't work with inspect, but inspect does the right thing.  So I wrote a horrible python wrapper for the cython function, with all the parameters.  Yay.

https://github.com/dwavesystems/minorminer/pull/86

PR #86, opened 2019-06-12
added inspect-friendly python wrapper to overcome Cython's opinions

boothby:
This is a counter-proposal to @joseppinilla's PR #82 -- it makes a wrapper around the cython function, which exposes all the parameters and provides defaults.  It's already a pain to add / modify parameters... so this nasty mess doesn't change the difficulty much.


python/minorminer.py
```diff
@@ -0,0 +1,42 @@
...
+                            skip_initialization=skip_initialization,
+                            verbose=verbose,
+                            initial_chains=initial_chains,
+                            fixed_chains=fixed_chains,
+                            restrict_chains=restrict_chains,
+                            suspend_chains=suspend_chains,
+                            )
+
+
+find_embedding.__doc__ = __find_embedding.__doc__
```
arcondello: I think https://docs.python.org/2/library/functools.html#functools.wraps will give you this plus cover some other edge cases.

python/minorminer.py
```diff
@@ -0,0 +1,42 @@
+from minorminer_c import miner, VARORDER, find_embedding as __find_embedding
+
+
+def find_embedding(S, T,
```
arcondello, line 7: I would just add a quick comment explaining why you're doing this

arcondello (approved):

python/minorminer.py
boothby: Is that nicer?  I find the following to be surprising/misleading:

`In [2]: minorminer.find_embedding`
`Out[2]: <function minorminer_c.find_embedding>`

arcondello: Fair point. Though it could be argued that for practical purposes it is that function in the `minorminer_c` module, you are functionally just importing it into the `minorminer` namespace. I am honestly unsure which is more surprising.

boothby: I made the change that produced that line and the code looks cleaner, so I think I'll just ship it as is despite the wart

PR #89, opened 2019-07-12
Fixes error types in minorminer.miner

joseppinilla:
I was having issues using miner without any parameters and using the quality_key method.

I suggest parsing "fixed_chains" before parsing "suspend_chains"

Also, the line `if "suspend_chains" in params:` no longer made sense, because the minorminer.py wrapper is adding default values to all parameters.


boothby (commented):
~~I like these changes.  Would you please add a minimal demonstration of the bug you're fixing, and add it to `tests/test_lib.py`?~~

Sorry, that's not a great request since I haven't taken the time to test the miner class.  I'll accept as is.

PR #93, opened 2019-10-29
Version 0.1.9

boothby:


setup.py
```diff
@@ -65,8 +65,8 @@ class Extension(extension.Extension, object):
 ext = '.pyx' if USE_CYTHON else '.cpp'
 
 extensions = [Extension(
-    name="minorminer_c",
-    sources=["./python/minorminer_c" + ext],
+    name="_minorminer",
```
arcondello, line 68: I think you just need to change this to `minorminer/_minorminer` to fix the tests

PR #94, opened 2019-10-29
Convert to a package as well as CI refresh

arcondello:
Includes the commits from and therefore closes #93

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -226,7 +226,7 @@ jobs:
   test-osx-3.7:
     <<: *osx-tests-template
     environment:
-      PYTHON: 3.7.5
+      PYTHON: 3.7.4
       HOMEBREW_NO_AUTO_UPDATE: 1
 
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -143,6 +143,11 @@ jobs:
             pip install cython
             pip install dist/minorminer-*.tar.gz
 
+  test-3.7:
+    <<: *full-test-template
+    docker:
+      - image: circleci/python:3.6-jessie
```
boothby: This doesn't look right

arcondello:
I would like to do a deeper refactor of the appveyor, but since a lot of it comes from python 2.7 and 3.4 and we're dropping support for those soon, it's probably best to wait until the new year.

boothby:
~~This is all looking great.  Is it trouble if I ask you to squash the commits?~~

Oh dear, I forgot the "squash and merge" option :roll_eyes:

PR #98, opened 2019-12-04
Features/layout initialization

aidanproy:
Provide an example of how the initial_chains option of find_embedding can be used to generate better chains.


boothby:
This is subsumed by the `layout` package.

PR #101, opened 2020-01-11
Layout and placement

stefanhannie:
Here is the initial spring layout implementation. At a high level it's dead simple:
- Compute the layouts for `S` and `T`
- Map vertices of `S` to the "closest" vertices of `T`
- Turn that mapping into `initial_chains`
- Hand those `initial_chains` to `minorminer`


minorminer/layout/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+
+from minorminer.layout.construction import *
+from minorminer.layout.hinting import *
+from minorminer.layout.layout import *
+from minorminer.layout.placement import *
+
+
+def find_embedding(S, T, layout=kamada_kawai, placement=closest, construction=singleton, hinting=initial, **kwargs):
+    """
+    Tries to embed S in T by computing layout aware initial_chains and passing them to minorminer.
```
JoelPasvolsky: initial_chains --> this can be made into a reference to the parameter, then users can click to it
"computing layout aware" should be "computing layout-aware"
"passing them to minorminer" --> we're in minorminer. Do you mean the previous find_embedding? Does this function need a different name to distinguish it from that one?

minorminer/layout/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Parameters
+    ----------
+    S : NetworkX graph or edges data structure (dict, list, ...)
+        The graph you are embedding (source) or a NetworkX supported data structure (see to_networkx_graph()).
+    T : NetworkX graph or edges data structure (dict, list, ...)
+        The graph you are embedding into (target) or a NetworkX supported data structure (see to_networkx_graph()).
+    layout : function or (function/dict, function/dict) (default kamada_kawai)
```
JoelPasvolsky: Will be cvlearer if you move the default to the end of the description instead. "S and the second applies to T. Default is kamada_kawai."

minorminer/layout/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+        Will either compute a layout or pass along precomputed layouts. If it's a single object, it has to be a function
```
JoelPasvolsky: "Will either compute a layout or pass along precomputed layouts." --> I think what you're saying here is that this parameter ""Specifies either a single function to compute the layout for both S and T or two pre-computed layouts, the first for S and the second for T."

minorminer/layout/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+        and it applies to both S and T. If it's a tuple of objects the first applies to S and the second applies to T.
+    placement : function (default closest)
+        The placement algorithm to call.
```
JoelPasvolsky: Could you add a bit more here?

minorminer/layout/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+        The chain construction algorithm to call.
+    hinting : function (default initial)
+        The type of minorminer hinting to call.
+    kwargs : dict 
+        Keyword arguments are passed to various functions.
+
+    Returns
+    -------
+    emb : dict
+        Output is dependant upon kwargs passed to minonminer, but more or less emb is a mapping from vertices of
```
JoelPasvolsky: dependent

minorminer/layout/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+        G = T
+    else:
+        G = nx.Graph(T)
+
+    return H, G
+
+
+def parse_kwargs(kwargs):
+    """
+    Pull out kwargs for layout and construction functions. Leave the remaining ones for minorminer.
```
JoelPasvolsky: Consider replacing "Pull out" with "Extract", we like to be fancy here

minorminer/layout/construction.py
```diff
@@ -0,0 +1,122 @@
+from collections import defaultdict
+
+import minorminer as mm
+
+
+def singleton(S, T, placement, extend=False):
+    """
+    Given a placement (a map phi from vertices of S to vertices of T), form the chain [phi(u)] for each u in S.
```
JoelPasvolsky: a map, ``phi``, from vertices

minorminer/layout/construction.py
```diff
@@ -0,0 +1,122 @@
...
+    chains = {u: [v] for u, v in placement.items()}
+    if extend:
+        return extend_chains(S, T, chains)
+
+    return chains
+
+
+def neighborhood(S, T, placement, second=False, extend=False):
+    """
+    Given a placement (a map phi from vertices of S to vertices of T), form the chain N_T(phi(u)) (closed neighborhood
```
JoelPasvolsky: a map, ``phi``, from vertices

minorminer/layout/placement.py
```diff
@@ -0,0 +1,60 @@
...
+    placement = {}
+    for u, u_pos in S_layout.items():
+        placement[u] = min(
+            T_layout, key=lambda v: euclidean(u_pos, T_layout[v]))
+    return placement
+
+
+def injective(S_layout, T_layout):
+    """
+    Injectively maps vertices of S to the closest vertices of T as given by S_layout and T_layout. i.e. This is the
```
JoelPasvolsky: "and T_layout. i.e. This is" --> either "and T_layout; i.e. this is" or else "and T_layout. This is"

JoelPasvolsky (approved):
LGTM. One thing that confuses me is why is the find_embedding() in file layout/__init__.py? That seems unusual to me. Is it temporary and then going through a renaming?

boothby:
> LGTM. One thing that confuses me is why is the find_embedding() in file layout/**init**.py? That seems unusual to me. Is it temporary and then going through a renaming?

This was my recommendation, actually -- it's not _terribly_ unusual but the function may move at a later date.

minorminer/layout/__init__.py
stefanhannie: Almost, it's a bit more general. It can be a single function, or a tuple with any combination of functions and layouts. i.e. `(f(), {})` is fine. What about:

"Specifies either a single function to compute the layout for both S and T or a 2-tuple consisting of functions or pre-computed layouts, the first applying to S and the second applying to T."

minorminer/layout/__init__.py
stefanhannie: * I'm okay renaming the function. @boothby what do you think about `layout_find_embedding()`? A bit long, but it should just be a temporary name.
* How do I make a reference to something like `initial_chains`?

minorminer/layout/__init__.py
stefanhannie: Are you not worried about breaking the convention that all other keyword arguments are 
```arg_name : arg_type (default default_arg)```?

minorminer/layout/__init__.py
boothby: I like the function having the same name.  It appears in a submodule for the purpose of differentiation.  As for additional parameters, I'd simply refer the reader to the documentation for minorminer.find_embedding.

minorminer/layout/construction.py
stefanhannie: Is the comment here to add: 
* commas (I should have done that)
* write `phi` (with back-ticks)
* :math: `phi` (use the `:math:` prefix thing I've seen in other places in the project)

minorminer/layout/__init__.py
JoelPasvolsky: I afterwards noticed that minorminer uses this different convention from other Ocean tools.

arcondello:
Looks like cmake build issue is probably related to https://github.com/google/googletest/issues/2711

PR #106, opened 2020-03-09
remove python 2.7 and 3.4 build/deploy

boothby:


arcondello:
Think you're missing a skip in the bash loop for manylinux deploy

boothby:
Aren't you supposed to be on vacation?  I'm finding your comment a little cryptic, but it can wait 'til you're back in town -- I won't be deploying this week.

randomir:
@boothby, manylinux image includes many python versions, I believe @arcondello is referring to something like [this](https://github.com/dwavesystems/dimod/blob/b780a4adb937fb55889baec4611612f93cea1721/.circleci/config.yml#L350).

boothby:
Aha, thanks!

Issue #62, opened 2018-10-11
Debug/error messages printed on stdout/err


randomir:
Python library errors are better raised as exceptions than printed to stdout/stderr. That way it's easier to actually handle them (or programmatically acknowledge them). For debug/error logging we could use the standard logging module.

To reproduce (one of many possible error messages):

```
>>> emb = minorminer.find_embedding({}, [])
never tried to embed anything; the qubit graph has no edges>>> 
```

Generalizes issue #28.

boothby:
addressed in #117

Issue #28, opened 2018-01-29
find_embedding prints "failed during initialization. embeddings may be invalid."


tdgoodrich:
Sometimes when `find_embedding` is called on inputs that are borderline un-embeddable, the following message is printed:

`failed during initialization. embeddings may be invalid.`

Consider raising an exception instead, so the code calling the function knows something is amiss. Otherwise, consider validating the returned embedding so that it is guaranteed to be valid.

boothby:
addressed in #117

PR #113, opened 2020-03-17
Stop printing errors to stderr

boothby:
See conversation in issue #28 and #62 -- there's a couple of annoying conditions where errors are printing to stderr rather than raising exceptions.  This should fix that.


boothby:
ummm... dupe

PR #118, opened 2020-03-27
Layout initial

stefanhannie:
A clean fresh start. Let's start with these two things:

- Layout class implementation
- Placement class implementation

### DIFF
--- a/minorminer/layout/__init__.py
+++ b/minorminer/layout/__init__.py
@@ -1,0 +1,133 @@
+import time
+
+import minorminer as mm
+import networkx as nx
+
+from .layout import Layout
+from .placement import Placement
+
+
+def find_embedding(
+    S,
+    T,
+    layout=None,
+    placement=None,
+    mm_hint_type="initial_chains",
+    return_layouts=False,
+    **kwargs
+):
+    """
+    Tries to embed S in T by computing layout-aware chains and passing them to minorminer.find_embedding(). Chains are 
+    passed as either initial_chains or suspend_chains (see documentation for minorminer.find_embedding to learn more).
+    Parameters
+    ----------
+    S : NetworkX graph or edges data structure (dict, list, ...)
+        The graph you are embedding (source) or a NetworkX supported data structure (see to_networkx_graph()).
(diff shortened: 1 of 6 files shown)


minorminer/layout/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+    S_layout, T_layout = _parse_layout_parameter(S, T, layout, layout_kwargs)
+
+    # Compute the placement (i.e. chains)
+    S_T_placement = Placement(
+        S_layout, T_layout, placement, **placement_kwargs)
+
+    end = time.process_time()
+    timeout = kwargs.get("timeout")
+    if timeout:
+        kwargs["timeout"] = timeout - (end - start)
```
boothby: if the remaining time is zero or negative, it would be good to raise a `TimeoutException` here

minorminer/layout/placement.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def _parse_layout(G_layout):
+    """
+    Ensures a Layout object was passed in and makes a copy to save in the Placement object.
+    """
+    if isinstance(G_layout, layout.Layout):
+        return layout.Layout(G_layout.G, G_layout.layout)
+
+    if isinstance(G_layout, dict):
```
boothby: Why check if it's a `dict`?  I think this error message is useful in either case...

minorminer/layout/tests/test_placement.py
```diff
@@ -0,0 +1,81 @@
...
+        Test some failure conditions
+        """
+        # Dimension mismatch failure
+        self.assertRaises(ValueError, mml.Placement,
+                          self.S_layout, self.C_layout_3)
+
+        # Layout input failure
+        self.assertRaises(TypeError, mml.Placement,
+                          nx.spectral_layout(self.S), self.C_layout)
+        self.assertRaises(TypeError, mml.Placement, (), self.C_layout)
```
boothby: I'd duplicate these to directly test _parse_layout

boothby (changes requested):

PR #119, opened 2020-03-27
Placement closest

stefanhannie:
The `closest` placement strategy is introduced. 

- It looks at the `k` nearest neighbors of a given point, and then selects one based on a greedy overlap minimization.
- It can also consider subgraphs of `T` as possible points to map to in the placement.


minorminer/layout/tests/__init__.py
```diff
@@ -0,0 +1,51 @@
+import unittest
```
boothby: not a huge fan of this being named `__init__.py` -- perhaps name it `common.py` or something?

minorminer/layout/placement.py
```diff
@@ -1,22 +1,142 @@
...
+    return cheapest_subset
+
+
 class Placement(abc.MutableMapping):
     def __init__(
         self,
         S_layout,
         T_layout,
         placement=None,
+        fill_T=False,
```
boothby: This is kinda bikeshedding... but can you think of a better name than `fill_T`?  I kinda feel like `scale_ratio` might be informative and flexible?  I'm thinking `scale_ratio=None` wouldn't rescale, `scale_ratio=1` would rescale S to fit T, and anything less than 1 would take up smaller portions of T...

boothby (changes requested):

minorminer/layout/tests/__init__.py
stefanhannie: I can move it to a file `common.py` but `unittest` can't discover tests unless there is an `__init__.py` (but a blank one is fine).

boothby: yeah, a blank one is good

boothby (approved):

PR #123, opened 2020-04-01
modernize the C++ by replacing macros with templates

boothby:
This is a bit of an experiment.  I'm interested in modernizing the C++, replacing preprocessor directives with templates.  The first commit seems to have a significant impact on performance (a hit of about 15% on a standard benchmark, which is unacceptable).  I'm posting this as a PR mostly to see what explodes on various platforms (I've already found some deficiencies in clang).

### DIFF
--- a/include/chain.hpp
+++ b/include/chain.hpp
@@ -4,16 +4,6 @@
 
 namespace find_embedding {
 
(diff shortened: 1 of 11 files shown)


boothby:
Stale, try again later.

PR #135, opened 2020-07-02
added license header to source files

boothby:
For posterity's sake:
```
while read -r extensions comment; do
    for file in `git ls-files | grep "\.$extensions\$"`; do
        date=`git log --diff-filter=A --follow --format=%aI -- $file | tail -n 1 | sed 's/\(....\).*/\1/'`
        if [ $date != "2020" ]; then
            date=`echo $date - 2020`;
        fi;

        cat<<EOT > tmpfile
$comment Copyright $date D-Wave Systems Inc.
$comment
$comment    Licensed under the Apache License, Version 2.0 (the "License");
$comment    you may not use this file except in compliance with the License.
$comment    You may obtain a copy of the License at
$comment
$comment        http://www.apache.org/licenses/LICENSE-2.0
$comment
$comment    Unless required by applicable law or agreed to in writing, software
$comment    distributed under the License is distributed on an "AS IS" BASIS,
$comment    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
$comment    See the License for the specific language governing permissions and
$comment    limitations under the License.

EOT
        if ! grep -q -i '\(Apache\|copyright\)' "$file"; then
            cat $file >> tmpfile            
            mv tmpfile $file
        fi
    done;
done <<'____HERE'
\(py\|pyx\|pxi\|pxd\)  #
\(c\|h\|cpp\|hpp\) //
m %
____HERE
```


randomir:
This is really useful. But since it's here for posterity...

- [`for f in $(..)` pattern is fundamentally flawed](http://mywiki.wooledge.org/BashPitfalls#for_f_in_.24.28ls_.2A.mp3.29), and you'll want to replace it with `read` iterating over `NUL`-terminated list of files. Something like this:
```
while IFS= read -r -d '' file; do
    # process "$file"
done < <(git ls-files -z)
```

- when in doubt, quote: [variables under test](http://mywiki.wooledge.org/BashPitfalls#A.5B_.24foo_.3D_.22bar.22_.5D) (e.g. `$date` above), [filenames](http://mywiki.wooledge.org/BashPitfalls#cp_.24file_.24target) (e.g. `cat $file` above), [command substitution](http://mywiki.wooledge.org/BashPitfalls#cd_.24.28dirname_.22.24f.22.29) (e.g. `$(git log ...)` above), etc;

- [legacy backtick syntax has multiple issues](https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03), so POSIX now officially recommends the `$(..)` form of command substitution;

- not critical, but `date=$(echo $date - 2020)` can be simplified to `date="$date - 2020"`, and `sed 's/\(....\).*/\1/'` can be replaced with `cut -d- -f1`;

- also, instead of hard-coding `2020`, you might want to use `date +%Y`;

- maybe not relevant here, but in general, simply prepending the license will break the hashbang (if it existed).

PR #141, opened 2020-07-10
Fix missing fasteners dependency

arcondello:


boothby:
safe to merge, see  #142

PR #145, opened 2020-07-10
density-based embedding algorithm dispatch, take 1

boothby:


.circleci/config.yml
```diff
@@ -134,6 +134,7 @@ jobs:
           command: |
             . env/bin/activate
             python setup.py build_ext --inplace
+            python setup.py install
```
arcondello, line 137: This is installing minorminer on your python path, which can lead to confusing test results when some tests run against the installed version and others against the version in the CWD.

boothby:
As-is, this implementation is a bad idea.  Abandoning for now.

PR #148, opened 2020-07-22
Busgraph unittesting

boothby:


arcondello:
Wondering if we should consider migrating `verify_embedding` "upstream" to minorminer?

boothby:
That makes *tons* of sense.  I was thinking about duplicating the old `find_clique_embedding` stuff anyway, both to test against and to provide support for chimera graphs with `t>8`.

boothby:
I'll look at moving `dwave.embedding` wholesale into `minorminer`, and patching up the imports to maintain backwards compatibility for everybody using `dwave.embedding`

arcondello:
IMO things like `embed_bqm`, `unembed_sampleset` and the chain break resolution methods make sense in `dwave-system` rather than in minorminer (not least because minorminer has no notion of a BQM/SampleSet). So I think that part of the `dwave.embedding` namespace makes sense to keep there.

boothby:
Ah, yes, agreed.  I hadn't looked into the details yet.   I'm thinking:

```
polynomialembedder.py
chimera.py
diagnostic.py
pegasus.py
exceptions.py
```

boothby:
Summary of private conversation: yes, but not on this ticket.

boothby:
Test failure is an issue in `dne.find_clique_embedding`

PR #155, opened 2020-09-16
fixed issue in Pegasus K_4 embeddings

boothby:
Reported by an internal user.  Thanks, user!


boothby:
intentional dupe served its purpose :( what's up, CircleCI?

PR #158, opened 2020-09-17
Fix build for AppVeyor

hhtong:


randomir (commented):
Looks like [the PATH fix](https://github.com/dwavesystems/dwave-neal/pull/83/) works here as well.

But I would really, _really_, like to understand why these are showing:
```
C:\Python35\lib\site-packages\wheel\bdist_wheel.py:78: RuntimeWarning: Config variable 'Py_DEBUG' is unset, Python ABI tag may be incorrect
  warn=(impl == 'cp')):
C:\Python35\lib\site-packages\wheel\bdist_wheel.py:83: RuntimeWarning: Config variable 'WITH_PYMALLOC' is unset, Python ABI tag may be incorrect
  sys.version_info < (3, 8)))
```
AFAICT, it should be harmless. But it would be good to know for sure before we release a new version of Ocean.

Btw, appveyor utility files can now also be removed (see https://github.com/dwavesystems/qbsolv/pull/152/).

boothby:
Is this only on Py3.5?  That's in EOL, can we drop it now?

arcondello:
@boothby, good point. Should we just be dropping 3.5 in Ocean generally? @randomir 

https://github.com/dwavesystems/dwave-ocean-sdk/issues/91

randomir:
No, @boothby, it's on all versions.

@arcondello, sure, I don't mind dropping 3.5.

randomir:
@hhtong, as long as you're fixing appveyor, can you also add tests for py38?

hhtong:
@randomir Will do that right now. Should I also be removing the tests for 3.5 then? Or wait

arcondello:
I would wait on 3.5, it's a bigger conversation.

randomir:
@hhtong, I would wait with py35. It's not solving anything, and it's not in our way right now.

PR #167, opened 2020-10-29
Run doxygen on full include dir and provide full C documentation in RTD

JoelPasvolsky:
Currently RTD generates and displays very partial C documentation:
image

This update runs doxygen on all files under /include and displays namespaces, classes, files, and structs:
image

### DIFF
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -771,7 +771,7 @@ WARN_LOGFILE           =
 # spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = ../include ../include/find_embedding ../include/busclique
+INPUT                  = ../include 
 
 # This tag can be used to specify the character encoding of the source files
 # that doxygen parses. Internally doxygen uses the UTF-8 encoding. Doxygen uses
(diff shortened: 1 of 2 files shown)


JoelPasvolsky:
Another option that is more elegant as long as we always want to build C docs for all include files is:
```
INPUT                  = ../include
...
RECURSIVE              = YES
```

JoelPasvolsky:
@boothby, if we want to go with this more or less, I've got it working now locally for the SDK too.

boothby:
The recursive option certainly seems easier to maintain.

JoelPasvolsky:
I changed to the recursive method. Doxyfile has an `EXCLUDE` option that can then be used if we do want to exclude some subdirs in the future.

JoelPasvolsky:
The AppVeyor failure is not related to this PR:
```
File "C:\Python35\lib\site-packages\dwave\embedding\chimera.py", line 109, in find_clique_embedding
    raise ValueError("cannot find a K{} embedding for given Chimera lattice".format(len(nodes)))
ValueError: cannot find a K13 embedding for given Chimera lattice
```

boothby:
yeah... ironic when you test against an older implementation as a reference and discover a bug in it... this is very low on my todo list right now but it's safe to ignore

arcondello (approved):

arcondello:
Is this ready to merge? @JoelPasvolsky @boothby

Issue #149, opened 2020-07-22
Absorb some code from dwave.embedding 


boothby:
On #148 @arcondello brought up the idea of relocating some code from `dwave.embedding` into `minorminer`.  On review, the relevant files would be:

```
polynomialembedder.py
chimera.py
diagnostic.py
pegasus.py
exceptions.py
```

boothby:
done in #170

PR #168, opened 2020-11-10
Support py39

randomir:


randomir:
Seems like `dwave-system` (`dwave.embedding`) _is required_ in `minorminer/tests/test_busclique.py`.

That's a problem because `dwave-system` requires `minorminer`, thus creating a cyclic dependency.

@boothby, thoughts? Extracting/replicating just the minimum required for testing here?

boothby:
Should be resolved by #149 but that's currently on a (far) back burner.  Is this a priority for you?

randomir:
Yup, pretty high. It's blocking a release of new Ocean with python 3.9 support.

How about we move those busclique tests to dwave-system, until #149 is implemented? (If minorminer is the core dependency of dwave-system, it makes sense to test integration of dwave-system with minorminer in the higher-level package.)

boothby:
#149 is resolved

randomir:
Thanks @boothby!

Issue #173, opened 2020-12-03
Fix appveyor failures


arcondello:
See https://github.com/dwavesystems/minorminer/issues/172#issuecomment-738339228_

randomir:
Closed in #174.

PR #174, opened 2020-12-04
Remove illegal usage of char* in embedding diagnostics

boothby:
See issue #173


randomir (approved):
This looks good to me.

@boothby, can we merge it, or you wanted to add something else here?

boothby:
Thanks for the reminder @randomir I was gonna plow through some of the other warnings that come up in build but this ticket is fine as is.

PR #181, opened 2021-04-26
Add doctest and link checks

JoelPasvolsky:


docs/requirements.txt
```diff
@@ -1,4 +1,4 @@
-breathe<4.13.0
+breathe>4.14.0
```
randomir: we should probably bound sphinx here as well

JoelPasvolsky: Done, thanks!

randomir (approved):

JoelPasvolsky:
Why is Py3.5 still being tested here?

randomir:
PR to drop py35 coming in 10.

randomir:
Once #183 is merged, I'll rebase this PR (unless you'd prefer to do that, @JoelPasvolsky?)

JoelPasvolsky:
Either option is okay by me, @randomir.

JoelPasvolsky:
@randomir, rebased

PR #190, opened 2021-11-02
Zephyr layout

boothby:


minorminer/layout/__init__.py
```diff
@@ -172,7 +172,7 @@ def _parse_layout_parameter(S, T, layout, layout_kwargs):
         T_layout = t_layout
     else:
         # Use the dnx_layout if possible
-        if T.graph.get("family") in ("chimera", "pegasus"):
+        if T.graph.get("family") in ("chimera", "pegasus", "zephyr"):
```
randomir, line 175: Perhaps create an `Enum` for known families? Should prevent typos.

boothby: I don't disagree, but to do that properly, we'd probably want to define that enum in `dwave_networkx`.  How important do you think this is?

randomir: Makes sense. I don't think it's critical, as in a blocker of this PR. I'll open an issue.

pau557: Slightly related; leaving this here so it doesn't get lost.

Chatting with Kelly we were considering the following unified scheme

```python
topology = dnx.Topology("zephyr")
g = topology.generator(16)
topology.plotter(g)
```

to avoid tedious `if` `elif` logic that tends to appear everywhere:

```python 
if topology == 'pegasus':
    generator = pegasus_graph
    plotter = draw_pegasus
elif topology == 'zephyr':
    generator = zephyr_graph
    plotter = draw_zephyr

g = generator(5)
plotter(g)
```

PR #192, opened 2021-11-12
Coordinate safety

boothby:
I found a bug while testing manually, where non-square Chimera segfaulted.  This codebase is very fussy because of the plurality of coordinate systems that are involved.  So, I bit the bullet and implemented some compile-time checking.  It was a bit of a ride.  Rather than rewrite it here, I'll paste in the comment from `coordinate_types.hpp` (which is roughly duplicated in the commit log)
```
Coordinate arithmetic is absolutely fraught with challenges.  In order to 
reduce the possibility of bugs, we've introduced an extremely pedantic system
of coordinate types.  The busclique algorithms use two kinds of coordinate 
systems:

  * matrix coordinates (y, x) which are featured in the chimera coordinate
    system (y, x, u, k) -- where u indicates the 'orientation' of a qubit,
    either vertical(u=0) or horizontal(u=1) and k indicates the 'shore index'
    of a qubit (irrelevant here)

  * relative coordinates (w, z) which are features in the pegasus and zephyr
    coordinate systems (u, w, k[, j], z) -- where u is the orientation as
    above and k and j are similarly irrelevant.

The relationship between matrix and relative coordinates is that

  * (y, x, u=0) is equivalent to (z=y, w=x, u=0), and

  * (y, x, u=1) is equivalent to (w=y, z=x, u=1)

We have encoded this equivalence in four types, size_y, size_x, size_w, and
size_z.  Each type supports arithmetic with itself, but forbids confusion.  For
example,

    size_y y0, y1;
    size_x x0, x0;
    (y0 + y1*y0 + y1)/(y1-y0) // cool
    y0 + x0;                  // compile error!

But, we need to convert between matrix and relative coordinates frequently!
We do that with the `vert` and `horz` functions:

    bool u = 0;
    size_y y;
    size_x x;
    size_w w0, w1;
    size_z z0, z1;
    w1 = w0 + vert(y);
    z1 = z0 + vert(x);
    u = 1;
    w1 = w0 + horz(x);
    z1 = z0 + horz(y);

Sometimes, we *really* need to mix things up.  There are two ways to convert
a coordinate to a size_t: implicitly through `coordinate_base::operator size_t`,
or explicitly through `coordinate_index()`.  The implicit conversion is
private, to prevent accidental implicit conversions.  The `coordinate_converter`
class is a friend to the coordinate types, and that's the only place we need to
be very careful.  We use the `coordinate_index()` method in very few one-off
instances.  Ideally, it should only be used for indexing into arrays:

    size_y y;
    uint8_t mem[100];
    mem[coordinate_index(y)];

This entire file, and the associated refactor, is a very heavy-handed approach
to fixing a bug (segfaulting non-square Chimera graphs).  However, it reduces a
significant amount of anxiety in developing and extending this library.  In 
small tests, code involving coordinate classes is optimized down to identical 
assembly to the same code using size_t in their place.  We'll see if that's true
in a complex codebase...

Update: Curious.  The performance was rather miserable on my system (gcc 9.3.0);
at around 3x slower than before the change that I hoped would be "zero cost."  I
decided to roll back very carefully; not changing any more code than necessary
to ensure a clean test.  I commented out the coordinate types, and replaced them
with `typedef size_t ...`.  In the initial version, the coordinate types had 
methods, specifically, `coordinate_base.index`, `coordinate_base.horz` and 
`coordinate_base.vert`.  I replaced these with global functions because we can't
attach methods to a size_t.  Sure enough, the performance returned to normal.
For sake of a fuller understanding, I refactored `coordinate_base` to support 
the same syntax, and lo and behold, the performance stayed the same.

Since I'm using a very up-to-date gcc, I am reluctant to put the coordinate
types into the release build.  If I was a docker shepherd, I'd probably spool up
a few dozen instances and benchmark on all the gcc's and clang's and msvc's that
I could get my hands on.  Alas, it is much easier to rely on the fact that CI
tests with debugging enabled, and retain the flag-dependent build since I've
already written it.  A note, I'd be oh-so-slightly more comfortable if `vert`,
`horz`, and `coordinate_index` were macros.  But I take a deep breath and remind
myself what century I'm programming in.```

### CODE REVIEW

File: include/busclique/coordinate_types.hpp [hunk 1]
```diff
@@ -0,0 +1,359 @@
...
+
+
+template<typename T>
+class coordinate_base;
+
+template<typename T>
+size_t coordinate_index(coordinate_base<T>);
+
+template<typename T>
+class coordinate_base {
```
hhtong, line 184: Would be nice to have docstrings for these classes and functions

include/busclique/util.hpp
```diff
@@ -150,57 +151,44 @@ class ignore_badmask {};
...
-    inline void chimera_coordinates(size_t q,
-                                    size_t &y, size_t &x, size_t &u, size_t &k) const {
-        k = q % shore;  q /= shore;
-        u = q % 2;      q /= 2;
-        x = q % dim[1]; y  = q/dim[1];
-        minorminer_assert(chimera_linear(y, x, u, k) == q);
+    template<typename shore_t>
+    inline void linear_chimera(size_t q, 
+                               size_y &y, size_x &x, bool &u, shore_t &k) const {
+        coordinate_converter::linear_chimera(q, y, x, u, k, dim_y, dim_x, shore_t(shore));
```
hhtong, line 201: For my own understanding: how come these functions aren't in chimera_spec_base instead?

include/busclique/coordinate_types.hpp
boothby: pushed :)

include/busclique/util.hpp
boothby: It's also used in `topo_cache`, `pegasus_spec_base` and `zephyr_spec_base` in handling of `badmask`

boothby:
This ticket finally solves all of the problems that I expected to be surfaced by introducing the coordinate system (assuming tests pass; they did locally).  I took a bit of a detour, in moving from `nose` to `pytest` because it made identifying problems easier.  I regret that it's turned into a bit of a patchbomb.  I could calve off the `pytest` migration to another ticket if anybody really hates that.

arcondello (approved):

test_requirements.txt
```diff
@@ -0,0 +1,2 @@
+pytest>=6.0.0
```
hhtong, line 1: Is there a reason for switching from `tests/requirements.txt` to `test_requirements.txt`?

boothby: yes, because the tests are in multiple directories (which, I'm not entirely prepared to defend, but I didn't want to make this patch even bigger)

boothby: at some point, I'll probably migrate to using nox (or, if that's unpopular, tox)

hhtong (approved):

Issue #197, opened 2022-01-18
Add support to release linux aarch64 wheels


odidev:
On aarch64, `pip install minorminer` builds the wheels from source code and then installs it. I can see that the job to build Linux aarch64 wheel is added in the config.yml file. May I know the plan for releasing Linux AArch64 wheels on PyPI.

arcondello:
Hi @odidev , just released [0.2.7.post0](https://github.com/dwavesystems/minorminer/releases/tag/0.2.7.post0).

PR #198, opened 2022-01-18
Add Linux aarch64 build to test and deploy

arcondello:
https://github.com/dwavesystems/minorminer/pull/196 added the infrastucture but it was not enabled.

See https://github.com/dwavesystems/minorminer/issues/197

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -316,6 +316,7 @@ workflows:
           matrix:
             parameters:
               python-version: &python-versions [3.6.8, 3.7.9, 3.8.9, 3.9.4, 3.10.0]
+      - build-linux-aarch64: *build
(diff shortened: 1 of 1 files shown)


randomir (approved):

randomir:
Note: one of our dependencies (`rectangle-packer`) [does not ship aarch64 wheel for py310](https://github.com/Penlect/rectangle-packer/issues/16). That's the last Ocean dependency without binary wheels for all platforms we support.

boothby:
Good point, Radomir.  I'll reach out to the `rectangle-packer` maintainer and request that wheel.

edit: :facepalm: shouldn't post before my coffee sets in, I see that your link is exactly that reach-out.

PR #199, opened 2022-01-18
Fix deploy regex in .circleci/config.yml

arcondello:


.circleci/config.yml
```diff
@@ -345,37 +345,21 @@ workflows:
     jobs:
       - build-linux:
           <<: *build
-          filters:
+          filters: &deploy-filters
             tags:
-              only: /^[0-9]+(\.[0-9]+)*((\.dev|rc)([0-9]+)?)?$/
+              only: /^[0-9]+(\.[0-9]+)*(\.(dev|rc|post)([0-9]+)?)?$/
```
randomir: If I'm reading [PEP-440](https://www.python.org/dev/peps/pep-0440/) correctly, normalized forms are:
```
rc
.post
.dev
```

All three start at zero, i.e. `rc == rc0`, etc.

In addition to `.`, post and dev can start with `-` and `_`, but they are normalized to `.`.

arcondello: Oh interesting. I thought that was an odd bug, should have checked it myself. Thanks! Will fix.

arcondello: Should be fixed.

randomir (approved):
LGTM

Issue #202, opened 2022-03-15
C++ Build Error


Gabriel-in-Toronto:
**Description**
I am trying to follow the guides and build the example.cpp and I am getting a error that 
```
object of abstract class type "MyCppInteractions" is not allowed:C/C++(322)
[example.cpp(32, 42): ]()pure virtual function "find_embedding::LocalInteraction::displayOutputImpl" has no overrider
[example.cpp(32, 42): ]()pure virtual function "find_embedding::LocalInteraction::displayErrorImpl" has no overrider
```
**To Reproduce**
Compile the example.cpp
 `g++ example.cpp -std=c++11 -o example -pthread` 
The output is
```
Starting build...
/usr/bin/g++ -fdiagnostics-color=always -g /home/minorminer/examples/example.cpp -o /home/minorminer/examples/example
/home/minorminer/examples/example.cpp: In function ‘int main()’:
/home/minorminer/examples/example.cpp:32:60: error: invalid new-expression of abstract class type ‘MyCppInteractions’
     params.localInteractionPtr.reset(new MyCppInteractions());
                                                            ^
/home/minorminer/examples/example.cpp:18:7: note:   because the following virtual functions are pure within ‘MyCppInteractions’:
 class MyCppInteractions : public find_embedding::LocalInteraction {
       ^~~~~~~~~~~~~~~~~
In file included from /home/minorminer/examples/../include/find_embedding/graph.hpp:21:0,
                 from /home/minorminer/examples/../include/find_embedding/find_embedding.hpp:23,
                 from /home/minorminer/examples/example.cpp:16:
/home/minorminer/examples/../include/find_embedding/util.hpp:111:18: note:     virtual void find_embedding::LocalInteraction::displayOutputImpl(int, const string&) const
     virtual void displayOutputImpl(int loglevel, const string&) const = 0;
                  ^~~~~~~~~~~~~~~~~
/home/minorminer/examples/../include/find_embedding/util.hpp:114:18: note:     virtual void find_embedding::LocalInteraction::displayErrorImpl(int, const string&) const
     virtual void displayErrorImpl(int loglevel, const string&) const = 0;
                  ^~~~~~~~~~~~~~~~

Build finished with error(s).
```

boothby:
@Gabriel-in-Toronto can I ask why you closed this?  I confirmed that it's an issue on my machine.  If you've got a workaround, I'd encourage you to submit a PR.  This is a low priority issue for me, but it is a bug.

[...]

PR #204, opened 2022-03-17
Close Issue #202

Gabriel-in-Toronto:
PR for closing Issue #202

### DIFF
--- a/include/find_embedding/find_embedding.hpp
+++ b/include/find_embedding/find_embedding.hpp
@@ -85,8 +85,8 @@ class parameter_processor {
 
     vector<int> _filter_fixed_vars() {
         vector<int> unscrew(num_vars);
-        assert(var_fixed_unscrewed.size() == num_vars);
-        assert(num_fixed < num_vars);
+        iszero(var_fixed_unscrewed.size() == num_vars);
+        iszero(num_fixed < num_vars);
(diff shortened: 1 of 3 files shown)


include/find_embedding/find_embedding.hpp
```diff
@@ -85,8 +85,8 @@ class parameter_processor {
 
     vector<int> _filter_fixed_vars() {
         vector<int> unscrew(num_vars);
-        assert(var_fixed_unscrewed.size() == num_vars);
-        assert(num_fixed < num_vars);
+        iszero(var_fixed_unscrewed.size() == num_vars);
```
boothby, line 88: This appears to be a source of problems.  Better to change this to `minorminer_assert`

include/find_embedding/util.hpp
```diff
@@ -94,10 +94,10 @@ class LocalInteraction {
   public:
     virtual ~LocalInteraction() {}
     //! Print a message through the local output method
-    void displayOutput(int loglevel, const string& msg) const { displayOutputImpl(loglevel, msg); }
+    void displayOutput(int loglevel, const string& msg) const { displayOutputImpl(msg); }
```
boothby, line 97: Please do not remove `loglevel` from these functions, as it's used elsewhere in the library.  It would be better to add this parameter into the `MyCppInteractions` overrides and ignore it there.

boothby (changes requested):

Issue #175, opened 2020-12-08
KeyboardInterrupt tests failing on py39, macOS


randomir:
Failing tests: `test_lib.test_headless_interrupt` and `test_lib.test_interactive_interrupt`.

See [test-osx-3.9](https://app.circleci.com/pipelines/github/dwavesystems/minorminer/135/workflows/e2f263ce-ee52-4541-ba3f-1d69e903d78a/jobs/5264).

randomir:
@boothby, would you prefer we keep shaving that yak (from 6dcee69702184526345246bcdc9644f807bacb47, [test_lib.py#L855-L884](https://github.com/dwavesystems/minorminer/blob/6dcee69702184526345246bcdc9644f807bacb47/tests/test_lib.py#L855-L884)), exclude osx-3.9, or remove kbd int tests altogether?

randomir:
To unblock the release, I patched these tests to skip on osx-py39 as well. That means we're now testing for keyboard interruptibility only on Linux and macOS py35-py37. 

This issue is to be closed when a proper fix is implemented.

boothby:
That's best for now.  I would like to get to the bottom of this, but I've been busy with design work.

PR #210, opened 2022-05-20
Change to enable running two unit tests on MacOS

seatim:
This PR makes a small change to a sleep time in one function that is used by two unit tests.  That change enables one unit test to pass that was otherwise failing on MacOS.  Changing the sleep time does not compromise the integrity of the test, I believe.  With all tests now passing on MacOS, the code to skip two tests is not needed, so that code is removed.

### DIFF
--- a/tests/test_lib.py
+++ b/tests/test_lib.py
@@ -858,8 +858,8 @@ def _long_running_successful_problem(interactive):
                             interactive=interactive, timeout=2)
     except KeyboardInterrupt:
         sys.exit(2)
-    if time.perf_counter() - t0 > .5:
-        # be a little generous here... but the caller should kill this in way less than .5s
+    if time.perf_counter() - t0 > 1.5:
+        # be a little generous here... but the caller should kill this in way less than 1.5s
         sys.exit(1)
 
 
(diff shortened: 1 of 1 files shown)


seatim:
Well, shoot.  The change to the timeout causes a failure on Linux, apparently.

tests/test_lib.py
```diff
@@ -904,7 +872,7 @@ def run_interactive_interrupt(interactive):
     p = multiprocessing.Process(
         target=_long_running_successful_problem, args=(interactive,))
     p.start()
-    time.sleep(.1)
+    time.sleep(1)
```
boothby: See line 858, you might bump the `_long_running_successful_problem` to time out after 2s

seatim: I made that change locally and tested just now on both Mac and Linux, and the tests still pass.

seatim: Added that to the PR.

boothby: Ah, this is a ridiculously complex test.  Line 893:
```python
    if time.perf_counter() - t0 > .5:
        # be a little generous here... but the caller should kill this in way less than .5s
        sys.exit(1)
```
I guess we should bump that to 1.5?

seatim: > Ah, this is a ridiculously complex test. 

😆 It sure is!

> I guess we should bump that to 1.5?

Tested that locally and it ~~is safe~~ seems safe and prudent.  Will do.

boothby:
This ticket closes #175.  Thanks for your help, @seatim.

Issue #222, opened 2022-08-17
mine_clique_embeddings arguments appear to be ignored


jackraymond:
**Description**
The behaviour of mine_clique_embeddings seems to be independent of the argument mask_bound, whereas we would expect the routine to become slower, and produce a higher quality cache (on average) when using large values.

**To Reproduce**
qpu = DWaveSampler()
G = qpu.to_networkx_graph()

BCquick = mine_clique_embeddings(G, mask_bound=1, heuristic_bound=0, num_seeds = 1)
BCslow = mine_clique_embeddings(G, mask_bound=2**8, heuristic_bound=0, num_seeds = 1) 
In this second, case we anticipate a much slower search (learding order linear in mask_bound), but instead it is independent of this argument.

Workaround (suggested by @boothby)

minorminer.busclique.busgraph_cache.clear_all_caches()
BCslow = mine_clique_embeddings(G, mask_bound=2**8, heuristic_bound=0, num_seeds = 1) 

**Expected behavior**
Runtime scales with mask_bound, quality of cache increases on average

**Environment:**
 - OS: [Ubuntu 16.04.4 LTS]
 - Python version: [e.g. 3.7.0]

**Additional context**

PR #223, opened 2022-10-13
Add topology_identifier method to busgraph_cache

boothby:
To support external logistics of embedding caches, we need a way of identifying a graph in a consistent manner.  We already have functionality for that on a single host, but it's a largeish binary blob.  For external use, we use the hexdigest of a sha256 of that binary blob.

Also contains a drive-by bugfix of #222


boothby:
@dwave-bnugmanov that failed test is not a result of this patch, it's known to be a little noisy

dwave-bnugmanov (approved):

PR #232, opened 2023-05-01
Extend networkx support beyond version 2.4

boothby:
I've tested locally with py3.8.6 and networkx 2.4, 2.8.6 and 3.1

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ numpy==1.22.3;python_version>="3.8"
 numpy==1.21.6;python_version<"3.8"
 scipy==1.8.0;python_version>="3.8"
 scipy==1.7.3;python_version<"3.8"
-networkx==2.4
+networkx>=2.4
 dwave-networkx>=0.8.11
 fasteners==0.15
 homebase==1.0.1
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -152,7 +152,15 @@ class Extension(extension.Extension, object):
 
 python_requires = '>=3.7'
 install_requires = [
-    "scipy", "networkx", "dwave-networkx>=0.8.10", "numpy", "fasteners", "homebase", "rectangle-packer>=2.0.1"
+    "numpy==1.22.3;python_version>=3.8",
```
arcondello, line 155: You want to make the versions in `install_requires` as loose as possible. You also don't need to switch based on Python version, `pip` should handle that.

requirements.txt
```diff
@@ -2,7 +2,7 @@ numpy==1.22.3;python_version>="3.8"
 numpy==1.21.6;python_version<"3.8"
 scipy==1.8.0;python_version>="3.8"
 scipy==1.7.3;python_version<"3.8"
-networkx==2.4
+networkx>=2.4
```
arcondello, line 5: conversely you want the versions in `requirements.txt` to be pinned for reproducibility. We should also pin `dwave-networkx` while we're at it.

If you want to test against a wide dependency range, we should do that in CI.

PR #233, opened 2023-05-16
Test dependency ranges in CI

arcondello:
We only test the min and latest version of the dependency ranges. We could be more fine grained, but until we run into a specific issue I think this suffices.

Closes https://github.com/dwavesystems/minorminer/pull/232 as redundant.


randomir (approved):
lgtm

boothby (approved):

PR #238, opened 2023-08-29
Lower the OSX deployment target and drop Python 3.7

arcondello:
Also use a more modern cibuildwheel.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -311,22 +311,17 @@ workflows:
       - build-and-test-linux: &build
           matrix:
             parameters:
-              python-version: &python-versions [3.7.9, 3.8.9, 3.9.4, 3.10.0, 3.11.0]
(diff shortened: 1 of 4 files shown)


randomir (approved):
LGTM.

In the last round of CI refresh (while dropping py37) I updated it to always run tests on latest python patch version, but that can be done in a separate PR.

arcondello:
It's actually a bit of a misleading degree of specificity. Because cibuildwheel handles the environment of both the building and the testing and we use [`~= python version`](https://github.com/dwavesystems/minorminer/blob/ce62e20797cf74ab90f75b42969e6e49100aee10/.circleci/config.yml#L145) so it pulls the latest patch. We probably could/should rework it a bit to make that clearer.

randomir:
Oh, that's nice. Didn't catch the `~=`. :tada:

PR #239, opened 2023-08-30
Support Cython 3 and Python 3.12

arcondello:


.circleci/config.yml
```diff
@@ -291,13 +291,13 @@ jobs:
           command: |
             python -m venv env
             . env/bin/activate
-            pip install --upgrade << parameters.dependency-versions >>
+            pip install --upgrade setuptools pip wheel
```
randomir: Ideally we would test without setuptools, as it's not available by default in 3.12. (And it's specified as a build dependency in pyproject.toml already).

randomir (approved):
LGTM

.circleci/config.yml
arcondello: Let's [see how it goes](https://github.com/dwavesystems/minorminer/pull/239/commits/aca884457b1bd1888b0d9ba88ceafd8159431050)

PR #240, opened 2023-12-08
Rectangle-packer module

thisac:
- Adds rectangle-packer as a submodule to minorminer
- Removes the rectangle-packer requirement from `requirements.txt`

> [!NOTE] 
> First attempted to add [rectangle-packer](https://github.com/Penlect/rectangle-packer) as a submodule to the [external](https://github.com/dwavesystems/minorminer/tree/main/external) folder, but the build failed due to required `.pyx`/`.pxd` files not being found in path (see [here](https://groups.google.com/g/cython-users/c/ILEwFY1FZBs) for details). It seems like the extension name in `setup.py` is required to follow the folder structure, which causes compilation issues when the `.pyx`/`.pxd` files are outside the `minorminer` source package folder.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,8 @@ class build_ext_compiler_check(build_ext):
 
         for ext in self.extensions:
             arg_key = ext.extra_compile_args
-            ext.extra_compile_args = extra_compile_args[arg_key][compiler]
+            if arg_key:
+                ext.extra_compile_args = extra_compile_args[arg_key][compiler]
 
         link_args = extra_link_args[compiler]
         for ext in self.extensions:
@@ -83,6 +84,7 @@ class Extension(extension.Extension, object):
 
 
 ext = '.pyx' if USE_CYTHON else '.cpp'
+ext_c = '.pyx' if USE_CYTHON else '.c'
 
 glasgow_cc = [
     '/'.join(['external/glasgow-subgraph-solver/src', f])
@@ -132,6 +134,12 @@ extensions = [
         language='c++',
         extra_compile_args = 'glasgow',
     ),
+    Extension(
+        name="minorminer._extern.rpack._core",
+        sources=["./minorminer/_extern/rpack/_core" + ext_c, "./minorminer/_extern/rpack/src/rpackcore.c"],
+        include_dirs=["./minorminer/_extern/rpack/include"],
+        language='c',
+    ),
 ]
 
 if USE_CYTHON:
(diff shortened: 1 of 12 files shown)


setup.py
```diff
@@ -173,6 +179,7 @@ class Extension(extension.Extension, object):
     packages=['minorminer',
               'minorminer.layout',
               'minorminer.utils',
+              'minorminer.rpack',
```
randomir: Since we're including rectangle-packer here verbatim (right?), I wouldn't expose it publicly like this. I'd keep it under `minorminer._vendor.rpack`, or similar. (`_vendor` namespace is used by Pip, but I've seen this handled differently by various packages over the years. E.g. `requests` used to pack a modified version of `urllib3` under `packages` namespace, etc.)

randomir (commented):
I can't tell from this PR, but if you actually modified the upstream `rpack` code in this PR, I'd suggest, for traceability, to split that into multiple commits:
1. import `rpack` snapshot (with clear indication of source/state/sha1 in commit message),
2. apply code changes in subsequent commits.

setup.py
arcondello: I am personally a fan of `extern/`, it's what we use in [dimod](https://github.com/dwavesystems/dimod). But flexible on the name. Some way of signalling that it's external to this package would be good.

arcondello:
You might consider using a git subtree for the import step. It means that we could make PRs back upstream with changes more easily.

thisac:
I went with copying over the relevant code for our usage (thought I might have to slightly alter it; I didn't), so the idea here was to have a separate version of rpack that we could maintain and alter to our needs if required.

> You might consider using a git subtree for the import step.

@arcondello Could definitely do that, although it makes us dependent on their repo again, right (not necessarily an issue, but I thought we wanted to move away from that).

> I'd suggest, for traceability, to split that into multiple commits

@randomir The rpack code isn't altered, other than a license bit at the top of each, but some (non-crucial) files from the original repo aren't included and others are slightly moved around. Would you still recommend doing this?

arcondello:
> Could definitely do that, although it makes us dependent on their repo again, right (not necessarily an issue, but I thought we wanted to move away from that).

We want to use our own build system and build timing etc. I don't mind the repos having their content linked in some way, either via a clear commit message, a git subtree, or a git submodule.

The advantage of subtree (rather than say a full submodule) is that the code is included directly in the git history.

I think any approach is reasonable though. They all have tradeoffs.

randomir:
> The rpack code isn't altered, other than a license bit at the top of each, but some (non-crucial) files from the original repo aren't included and others are slightly moved around. Would you still recommend doing this?

Absolutely. It's much easier to do `git diff` than to play detective inspecting manually what changed.

arcondello:
I am fine with this approach, though we should document in this PR some of the other approaches we tried, for future references.

arcondello (approved):

randomir (approved):
LGTM!

thisac:
> I am fine with this approach, though we should document in this PR some of the other approaches we tried, for future references.

Already added a note in the description box at the top, which I think summarizes the reasons for the approach we took here.

arcondello:
So you did, I just have reading comprehension issues :flushed:

PR #244, opened 2024-05-29
Layout bug fix

mahdiehmalekian:
In `layout.py`:

- `_graph_distance_matrix` is modified so that it consistently returns symmetric distance matrix.
- `dnx_layout` is ensured to behave as expected (at the moment it gives values out of range due to some bug in `dnx.zephyr_layout` and `dnx.pegasus_layout`).
- `_nx_to_dnx_layout` is modified so that it returns center[dim] on dimensions dim>=3.
- edges in`_rotate_to_minimize_area` is modified to contain all edges of the convex hull (used to miss one). In the same function, the returned value (points) is modified so that it actually returns the rotation[best_idx].points.

In `placement.py`:
- `_minimize_overlap` now avoids an avoidable OverFlowError

I've added a couple of tests to the test suite to detect these issues.

### DIFF
--- a/minorminer/layout/layout.py
+++ b/minorminer/layout/layout.py
@@ -161,19 +161,18 @@ def _graph_distance_matrix(G,
     if disconnected_distance is None:
         disconnected_distance = len(G)
 
-    return np.array(
-        [
-            [V.get(v, disconnected_distance) for v in sorted(G)]
-                 for u, V in all_pairs_shortest_path_length
-        ]
-    )
+    dist_list = [
+        (u, [V.get(v, disconnected_distance) for v in sorted(G)])
+         for u, V in nx.all_pairs_shortest_path_length(G)
+         ]
+    return(np.array([_[1] for _ in sorted(dist_list)]))
 
 def _p_norm_objective(layout_vector, G_distances, dim, p):
     """Compute the sum of differences squared between the p-norm and the graph 
     distance as well as the gradient.
 
     Args:
-        layout (Numpy Array):
+        layout_vector (Numpy Array):
             A vector indexed by sorted vertices of G whose values are points in 
             some metric space.
 
@@ -238,13 +237,13 @@ def _p_norm_objective(layout_vector, G_distances, dim, p):
 
 
 def dnx_layout(G, dim=None, center=None, scale=None, **kwargs):
-    """The Chimera or Pegasus layout from `dwave_networkx` centered at the origin
+    """The Chimera or Pegasus or Zephyr layout from `dwave_networkx` centered at the origin
     with ``scale`` as a function of the number of rows or columns. Note: As per 
     the implementation of `dnx.*_layout`, if :math:`dim>2`, coordinates beyond 
     the second are 0.
 
     By default, :func:`dnx_layout` is used to compute the layout for target 
-    Chimera or Pegasus graphs when :func:`minorminer.layout.find_embedding` is 
+    Chimera or Pegasus or Zephyr graphs when :func:`minorminer.layout.find_embedding` is 
     called.
 
     Args:
@@ -268,7 +267,7 @@ def dnx_layout(G, dim=None, center=None, scale=None, **kwargs):
 
     Returns:
         dict: :attr:`.Layout.layout`, a mapping from vertices of ``G`` (keys) to 
-        points in :math:`R^d` (values).
+        points in :math:`[-scale, scale]x[-scale, scale]x{0}^{d-2}+center` (values).
     
     Examples:
         This example creates a :class:`.Layout` object for a Pegasus graph, with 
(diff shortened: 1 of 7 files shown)


minorminer/layout/layout.py
```diff
@@ -161,19 +161,18 @@ def _graph_distance_matrix(G,
...
 
-    return np.array(
-        [
-            [V.get(v, disconnected_distance) for v in sorted(G)]
-                 for u, V in all_pairs_shortest_path_length
-        ]
-    )
+    dist_list = [
+        (u, [V.get(v, disconnected_distance) for v in sorted(G)])
+         for u, V in nx.all_pairs_shortest_path_length(G)
```
randomir: why not `all_pairs_shortest_path_length`?

minorminer/layout/layout.py
```diff
@@ -161,19 +161,18 @@ def _graph_distance_matrix(G,
...
+         ]
+    return(np.array([_[1] for _ in sorted(dist_list)]))
```
randomir: - missing space
- extra parenthesis
- `_` should be used for throwaway variables only, not here

We try to follow [pep-8](https://peps.python.org/pep-0008/) in Ocean.

minorminer/layout/layout.py
```diff
@@ -306,7 +305,19 @@ def dnx_layout(G, dim=None, center=None, scale=None, **kwargs):
     elif family == "zephyr":
         dnx_layout = dnx.zephyr_layout(
             G, dim=dim, center=dnx_center, scale=dnx_scale)
-
+    
+    # if the output of dnx_layout is not what it should be (at the moment there is a bug in dnx.zephyr_layout and dnx.pegasus_layout)
```
randomir: should link to github issue, if we have one; if not we should create it

tests/layout/test_placement.py
```diff
@@ -17,9 +17,10 @@
 
 import dwave_networkx as dnx
 import minorminer.layout as mml
-import networkx as nx
+import networkx as nx
```
randomir: trailing space

tests/layout/test_placement.py
```diff
@@ -168,9 +169,20 @@ def test_failures(self):
         # Layout input failure
         self.assertRaises(TypeError, mml.Placement,
                           "not a layout", self.C_layout)
-        self.assertRaises(TypeError, _parse_layout, "not a layout")
+        self.assertRaises(TypeError, _parse_layout, "not a layout")
```
randomir: trailing space

randomir (commented):
The code looks good (minus some minor styling issues), but I hope @boothby can check for algorithm correctness.

minorminer/layout/layout.py
```diff
@@ -161,19 +161,18 @@ def _graph_distance_matrix(G,
...
         disconnected_distance = len(G)
 
-    return np.array(
-        [
-            [V.get(v, disconnected_distance) for v in sorted(G)]
-                 for u, V in all_pairs_shortest_path_length
-        ]
-    )
+    dist_list = [
+        (u, [V.get(v, disconnected_distance) for v in sorted(G)])
```
boothby: The bug that you're addressing is due to sorting `G` -- node labels are not guaranteed to be comparable, so calling `sorted` here may raise an exception.  The good news is that iteration over either `G` or `all_pairs_shortest_path_length` should occur in the same order (barring a bug in `networkx`).

minorminer/layout/layout.py
boothby: This should be much simpler once sorting is removed.

minorminer/layout/layout.py
mahdiehmalekian: I have a PR for it [here](https://github.com/dwavesystems/dwave-networkx/pull/236). Should also create an issue for it?

mahdiehmalekian: Just created the issue [here](https://github.com/dwavesystems/dwave-networkx/issues/239). There is also a PR to resolve it [here.](https://github.com/dwavesystems/dwave-networkx/pull/236)

minorminer/layout/layout.py
mahdiehmalekian: Because in case `G` is disconnected, `nx.all_pairs_shortest_path_length` doesn't return a value for nodes in different connected components. So, in that case, if we use `V.get(v)`, we'll end up with None values in the returned array, whereas for the purposes of the algorithm, it makes more sense to have a distance larger than the max distance in connected components.

minorminer/layout/layout.py
mahdiehmalekian: Done in commit `efb48f1`.

minorminer/layout/layout.py
mahdiehmalekian: Done in commit efb48f1.

tests/layout/test_placement.py
mahdiehmalekian: Fixed in commit efb48f1.

tests/layout/test_placement.py
mahdiehmalekian: Fixed in commit efb48f1.

minorminer/layout/layout.py
mahdiehmalekian: Done in commit `efb48f1`.

minorminer/layout/layout.py
```diff
@@ -161,19 +161,18 @@ def _graph_distance_matrix(G,
     if disconnected_distance is None:
         disconnected_distance = len(G)
 
-    return np.array(
-        [
-            [V.get(v, disconnected_distance) for v in sorted(G)]
-                 for u, V in all_pairs_shortest_path_length
-        ]
-    )
+    dist_list = [
```
boothby: You can do this even more simply:
```python
return np.array([
    [V.get(v, disconnected_distance) for v in G]
    for _, V in nx.all_pairs_shortest_path_length(G)
])
```

mahdiehmalekian: Done in commit `43b49c6`.

mahdiehmalekian:
Commit `de2d5dd`:

In `layout.py`, the algorithm for `_rotate_to_minimize_area` was changed. It now returns the rotated points with the minimum area of bounding parallel-to-axes box.

mahdiehmalekian:
@randomir Is this ready to be merged?

randomir (approved):
Looks good to me.

@boothby, would you like to take another (final) look before we merge?

mahdiehmalekian:
Yes, it's enough to fix it upstream in dwave-networkx, and then require a
version of dnx that's fixed.

tests/layout/test_placement.py
```diff
@@ -174,3 +175,14 @@ def test_failures(self):
...
+def test_minimize_overlap(): # To test whether an avoidable OverFlowError in _minimize_overlap is skipped
+    dim_a = 10
+    dim_b = 55
+    grid_edges = list(nx.grid_graph([dim_a, dim_b]).edges())
+    diagonal_edges = [((i, j), (i+1, j+1)) for i in range(dim_b-1) for j in range(dim_a-1)] + [((i, j), (i-1, j+1)) for i in range(1, dim_b) for j in range(dim_a-1)]
+    king = nx.Graph(grid_edges+diagonal_edges)
+    T = dnx.zephyr_graph(15)
+    mml.find_embedding(king, T, scale=1)
+    
+test_minimize_overlap()
```
randomir: I don't think this line is needed, the test should be picked up by pytest, no?
```suggestion
```

mahdiehmalekian: Sure, done in `01bec47`

PR #248, opened 2024-07-08
Use M1 resource class for macos builds on CircleCI

arcondello:
Also remove the (long outdated) coveralls tests in CI. We can consider adding it back in in a followup PR.

### DIFF
--- a/tests/test_lib.py
+++ b/tests/test_lib.py
@@ -839,13 +839,13 @@ class TestFindEmbedding(unittest.TestCase):
         return find_embedding(K, C, tries=1, chainlength_patience=0)
 
     @staticmethod
-    @unittest.skipIf(sys.platform == 'win32', 'Test may hang on win32, skipping out of caution')
+    @unittest.skipIf(os.getenv("CIRCLECI", None), 'Test is unreliable in CI')
     @success_perfect(1)
     def test_interactive_interrupt():
         return run_interactive_interrupt(True) == 0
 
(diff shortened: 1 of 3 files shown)


tests/test_lib.py
```diff
@@ -839,13 +839,13 @@ def test_variable_components_many():
         return find_embedding(K, C, tries=1, chainlength_patience=0)
 
     @staticmethod
-    @unittest.skipIf(sys.platform == 'win32', 'Test may hang on win32, skipping out of caution')
+    @unittest.skipIf(os.getenv("CIRCLECI", None), 'Test is unreliable in CI')
```
thisac, line 842: Didn't know that you could skip tests in CircleCI this way.

thisac (approved):

tests/test_lib.py
arcondello: Yeah, they have some [built-in environment variables](https://circleci.com/docs/variables/#built-in-environment-variables) for this purpose.

randomir: Which kind of defeats the purpose of CI, but I feel the pain. I could never make neal/greedy/tabu/hybrid concurrency tests work reliably on AppVeyor. :disappointed:

arcondello: This test probably could/should be rewritten to be a bit more reliable. it relies on timing the interrupt. But I was trying to minimize changes here.

PR #253, opened 2024-11-06
updated lib to cpp20

Willu12:
Due to changes in template functions, pairing_queue.hpp and util.hpp needed tweaking in order to work with cpp20


arcondello:
Hi @Willu12 can you say a bit more about what was failing before? With the exception of the
```
    pairing_node<N>(Args... args) : N(args...), next(nullptr), desc(nullptr) {}
```
```
    pairing_node(Args... args) : N(args...), next(nullptr), desc(nullptr) {}
```
and the `inline` change most of these changes seem to be stylistic?

PR #256, opened 2024-12-03
Update package structure

thisac:
An effort to make minorminer more Ocean-like.
* Removes Matlab support (deprecated since 0.2.8)
* Removes unnecessary function imports
* Remove git hook scripts
* Update package structure to be more like the rest of Ocean
* Move `test_requirements.txt` to `tests/requirements.txt`


pyproject.toml
```diff
@@ -7,9 +7,12 @@ requires = [
 build-backend = "setuptools.build_meta"
 
 [project]
+name = "minorminer"
+dynamic = ["version"]
 authors = [
-    {name = "Kelly Boothby", email = "boothby@dwavesys.com"},
```
arcondello, line 11: IMO we should do
```.toml
authors = [
    {name = "Kelly Boothby"},
]
maintainers = [
    {name = "D-Wave Inc.", email = "tools@dwavesys.com"},
]
```

arcondello (commented):
I missed that we started using `reno` in this package. We should add it to the `requirements.txt` and to the README under contributing.

releasenotes/notes/remove-matlab-support-61797b65ff5bfa5a.yaml
```diff
@@ -0,0 +1,3 @@
+---
+upgrade:
+  - Matlab support was deprecated as of vesion 0.2.8 and removed in the current version.
```
randomir:
```suggestion
  - Matlab support was deprecated as of version 0.2.8 and removed in the current version.
```

randomir (approved):
LGTM

pyproject.toml
thisac: Happy to go with that; makes sense to me.

arcondello (approved):

PR #259, opened 2024-12-27
parallel_embeddings.py and feasibility.py minor corrections

jackraymond:
Minor improvements to new parallel_embeddings.py and feasibility.py modules:
- max_num_emb should be optional, allowing a None value that indicates no bound on the number of embeddings. This has been implemented.
- The lattice_size() function is merely a helper function for find_sublattice_embeddings() and doesn't have much generalization power beyond this. For this reason I have moved it into the parallel_embeddings.py module. Separately, a new non-trivial upper-bound function could be added to feasibility.py as a feature enhancement (noted in comments). 
- A timeout parameter is added allowing control of the total operation time.
- Tiling functionality is added. A use can provide a known embedding (with respect to a default or known tile) and this is then stenciled in place of a search.


minorminer/utils/parallel_embeddings.py
```diff
@@ -197,13 +199,33 @@ def find_multiple_embeddings(
     return embs
 
 
+def lattice_size(T: nx.Graph = None) -> int:
```
randomir: This signature allows for a parameterless call, which fails with an `AttributeError` exception.

minorminer/utils/parallel_embeddings.py
```diff
@@ -128,7 +128,7 @@ def find_multiple_embeddings(
         S: The source graph to embed.
         T: The target graph in which to embed.
         max_num_emb: Maximum number of embeddings to find.
-            Defaults to 1, set to float('Inf') to find the maximum possible
+            Defaults to 1, set to None to find the maximum possible
```
randomir, line 164: This is actually a breaking change. For backwards compatibility, consider supporting `inf` as well, but throwing a deprecation warning if called with it. (I haven't checked, but if `Inf` support hasn't been released yet, than no need to deprecate it, obviously.)

minorminer/utils/parallel_embeddings.py
```diff
@@ -246,7 +268,8 @@ def find_sublattice_embeddings(
            family matching T). ``lattice_size_lower_bound()``
            provides a lower bound based on a fast feasibility filter.
         max_num_emb: Maximum number of embeddings to find.
-            Defaults to inf (unbounded).
+            Defaults to 1, set to None for unbounded (try unlimited search on
```
randomir, line 271: Backwards-compat comment from above seems to apply here as well.

minorminer/utils/parallel_embeddings.py
randomir: Indeed, this code hasn't been released yet, so please ignore the comment above.

jackraymond: The first pull request was merged prematurely. I think @thisac wants to take a second look at this tidy up request as well.

minorminer/utils/parallel_embeddings.py
```diff
@@ -101,17 +103,46 @@ def embeddings_to_array(embs: list, node_order=None, as_ndarray=False):
         return [[emb[v] for v in node_order] for emb in embs]
 
 
+def array_to_embeddings(embs: list, node_order=None):
```
randomir:
```suggestion
def array_to_embeddings(embs: list, node_order: Optional[Iterable] = None) -> list[dict]:
```
with
```python
from collections.abc import Iterable
```
added to imports.

minorminer/utils/parallel_embeddings.py
```diff
@@ -101,17 +103,46 @@ def embeddings_to_array(embs: list, node_order=None, as_ndarray=False):
...
+            variables in each row. When not provided variables are ordered to
+            match the first embedding :code:``embs[0].keys()``. node_order
+            can define any subset of the source graph nodes (embedding
+            keys).
+
+    Returns:
+        An embedding dictionary
+    """
+    if len(embs) is None:
+        return []
```
randomir: `len()` never evaluates to `None`.
```suggestion
    if not embs:
        return []
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -101,17 +103,46 @@ def embeddings_to_array(embs: list, node_order=None, as_ndarray=False):
...
+        embs: A list of embeddings, each embedding is a list where values are
+            chains in node order.
+        node_order: An iterable giving the ordering of
+            variables in each row. When not provided variables are ordered to
+            match the first embedding :code:``embs[0].keys()``. node_order
+            can define any subset of the source graph nodes (embedding
+            keys).
+
+    Returns:
+        An embedding dictionary
```
randomir: Looks like this function returns a list of dictionaries.

minorminer/utils/parallel_embeddings.py
randomir: I don't see this function used anywhere? 

Consider removing. If you want to keep it here, it would be good to add some tests.

minorminer/utils/parallel_embeddings.py
```diff
@@ -265,13 +411,20 @@ def find_sublattice_embeddings(
...
 
     Raises:
         ValueError: If the target graph ``T`` is not of type zephyr, pegasus, or chimera.
+            If the tile_embedding is incompatible with the source graph and tile.
 
     Returns:
         list: A list of disjoint embeddings.
     """
+
+    timeout = perf_counter() + timeout
```
randomir: Changing input parameter meaning (relative -> absolute time) could be a source of bugs and confusion. Rather, call it something like:
```suggestion
    timeout_at = perf_counter() + timeout
```

minorminer/utils/parallel_embeddings.py
randomir: In the meantime, a version that accepts `inf` was released, so we should now test for `inf`, and treat it the same way as `None`. Optionally, you could throw a warning in that case like:
```python
if max_num_emb == float('inf'):
    warnings.warn('Use of Inf for "max_num_emb" has been deprecated in favor of "None".',
                  DeprecationWarning, stacklevel=2)
    max_num_emb = None
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -197,20 +240,113 @@ def find_multiple_embeddings(
     return embs
 
 
+def lattice_size(T: Optional[nx.Graph] = None) -> int:
```
randomir: `T` doesn't look optional, but required.
```suggestion
def lattice_size(T: nx.Graph) -> int:
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -197,20 +240,113 @@ def find_multiple_embeddings(
...
+    return True
+
+def _is_valid_embedding(emb: dict, S: dict, T: dict, one_to_iterable: bool = True):
+    """If dwave.embedding module available check embedding validity. 
+
+    With special handling of 1:1 mappings."""
+    try:
+        from dwave.embedding import is_valid_embedding
+    except ImportError:
+        is_valid_embedding = _is_valid_embedding_import_failover
```
randomir: Replace with a module-level import
```suggestion
    from minorminer.utils.diagnostic import is_valid_embedding
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -197,20 +240,113 @@ def find_multiple_embeddings(
...
+    yielded = nx.Graph()
+    for p, q in target.subgraph(labels).edges():
+        yielded.add_edges_from((x, y) for x in labels[p] for y in labels[q])
+
+    for x, y in source.edges():
+        if x == y:
+            continue
+        if x in embedded and y in embedded and not yielded.has_edge(x, y):
+            return False
+    return True
```
randomir: Not needed, as `is_valid_embedding` can be imported from `minorminer.utils`.
```suggestion
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -17,10 +17,12 @@
 """
 import warnings
 
+from time import perf_counter
+import time
 import dwave_networkx as dnx
 import networkx as nx
 import numpy as np
-from typing import Union
+from typing import Union, Optional, Callable
```
randomir: Import order is: built-in/standard, 3rd-party, local.
```python
import time
from time import perf_counter
from typing import Union, Optional, Callable

import dwave_networkx as dnx
import networkx as nx
import numpy as np
```

randomir (approved):
LGTM.

minorminer/utils/parallel_embeddings.py
jackraymond: It's useful to be able to invert embeddings_to_array. I've been using this internally. With respect to the code we are developing for lattice experiments its useful.

jackraymond: I'll add a test

randomir: Ok, thanks.

jackraymond:
Test failures seem to be unrelated to the pull request:
FAILED ../../../../../../../../../Users/distiller/project/tests/test_lib.py::TestFindEmbedding::test_fail_timeout - AssertionError: test fail
I think I fully implemented the review recommendations

minorminer/utils/parallel_embeddings.py
```diff
@@ -147,13 +182,17 @@ def find_multiple_embeddings(
             diversification of the embeddings found.
         seed: seed for the ``numpy.random.Generator`` controlling shuffling (if
             invoked).
+        timeout: total time allowed across all embeddings in seconds. Time elapsed is
```
thisac:
```suggestion
        timeout: Total time allowed across all embeddings in seconds. Time elapsed is
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -166,8 +205,20 @@ def find_multiple_embeddings(
         _T = shuffle_graph(T, seed=prng)
     else:
         _T = T
+    if max_num_emb == float("inf"):
```
thisac:
```suggestion
        _T = T

    if max_num_emb == float("inf"):
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -181,7 +232,12 @@ def find_multiple_embeddings(
         ):
             emb = []
         else:
-            emb = embedder(_S, _T, **embedder_kwargs)
+            if perf_counter() >= timeout:
+                emb = []
+            else:
+                if timeout == 0:
+                    raise ValueError()
```
thisac: This check seems more suited to put at the top of the file. 

Also, should include a message in the error.

minorminer/utils/parallel_embeddings.py
```diff
@@ -181,7 +232,12 @@ def find_multiple_embeddings(
         ):
             emb = []
         else:
-            emb = embedder(_S, _T, **embedder_kwargs)
+            if perf_counter() >= timeout:
+                emb = []
```
thisac: I think it's clearer just to have break here as well as above.
```suggestion
                break
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -181,7 +232,12 @@ def find_multiple_embeddings(
...
+            else:
+                if timeout == 0:
+                    raise ValueError()
+                emb = embedder(_S, _T, **embedder_kwargs)
```
thisac, line 244: I'm assuming that the embedder can return `[]` and thus the break below is required?

minorminer/utils/parallel_embeddings.py
```diff
@@ -197,20 +253,63 @@ def find_multiple_embeddings(
...
+
+    The lattice size is the parameter ``m`` of a dwave_networkx graph, also
+    called number of rows, or in the case of a chimera graph max(m,n). This
+    upper bounds the ``sublattice_size`` for ``find_sublattice_embeddings``.
+
+    Args:
+        T: The target graph in which to embed. The graph must be of type
+            zephyr, pegasus or chimera and constructed by dwave_networkx.
+    Returns:
+        int: The maximum possible size of a tile
```
thisac:
```suggestion
        int: The maximum possible size of a tile.
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -197,20 +253,63 @@ def find_multiple_embeddings(
...
+    # Possible feature enhancement, determine a stronger upper bound akin
+    # to lattice_size_lower_bound, accounting for defects, the
+    # degree distribution and other simple properties.
+
+    return max(T.graph.get("rows"), T.graph.get("columns"))
+
+
+
+def _is_valid_embedding(emb: dict, S: dict, T: dict, one_to_iterable: bool = True):
+    """If dwave.embedding module available check embedding validity.
```
thisac:
```suggestion
    """Check embedding validity.
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -197,20 +253,63 @@ def find_multiple_embeddings(
...
+
+    With special handling of 1:1 mappings."""
+    from minorminer.utils.diagnostic import is_valid_embedding
```
thisac, line 284: Is there a reason for not putting this at the top of the file?

minorminer/utils/parallel_embeddings.py
```diff
@@ -197,20 +253,63 @@ def find_multiple_embeddings(
...
+
+    if one_to_iterable:
+        return is_valid_embedding(emb, S, T)
+    else:
+        return is_valid_embedding({k: (v,) for k, v in emb.items()}, S, T)
+
+
+def _mapped_proposal(emb: dict, f: Callable, one_to_iterable: bool = True):
+    if one_to_iterable:
```
thisac, line 293: Single line docstring would be nice to have even if hidden.

minorminer/utils/parallel_embeddings.py
```diff
@@ -265,13 +374,20 @@ def find_sublattice_embeddings(
             for diversification of the embeddings found.
         seed: seed for the `numpy.random.Generator` controlling shuffling (if
             invoked).
+        timeout: total time allowed across all embeddings in seconds. Time elapsed is
```
thisac:
```suggestion
        timeout: Total time allowed across all embeddings in seconds. Time elapsed is
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -265,13 +374,20 @@ def find_sublattice_embeddings(
...
 
     Raises:
         ValueError: If the target graph ``T`` is not of type zephyr, pegasus, or chimera.
+            If the tile_embedding is incompatible with the source graph and tile.
 
     Returns:
         list: A list of disjoint embeddings.
     """
+
+    timeout_at = perf_counter() + timeout
```
thisac: There's no check that `timeout > 0` here as above. For consistency maybe add it here as well?

minorminer/utils/parallel_embeddings.py
```diff
@@ -333,8 +449,52 @@ def find_sublattice_embeddings(
             "source graphs must a graph constructed by "
             "dwave_networkx as chimera, pegasus or zephyr type"
         )
+    if tile_embedding is not None:
+        if not _is_valid_embedding(tile_embedding, S, tile, one_to_iterable):
+            raise ValueError("tile_embedding is invalid for S and tile")
+        else:
+            use_filter = False  # Unnecessary
```
thisac: Minor, but else not required here.
```suggestion

        use_filter = False  # Unnecessary
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -333,8 +449,52 @@ def find_sublattice_embeddings(
...
+        # a subgraph. Can be violated given heuristic search
+        defect_free_embs = find_multiple_embeddings(
+            S,
+            tile,
+            max_num_emb=1,
+            use_filter=False,
+            seed=seed,
+            embedder=embedder,
+            embedder_kwargs=embedder_kwargs,
+            timeout=timeout_at - perf_counter(),
```
thisac, line 484: Could this cause issues if timeout is too small (but still >0) and thus `timeout_at - perf_counter()` becomes <= 0?

Perhaps worth adding a `if timeout_at - perf_counter() <= 1` directly above or changing this argument to `timeout=max(timeout_at - perf_counter(), 1)`, or something relevant?

minorminer/utils/parallel_embeddings.py
```diff
@@ -333,8 +449,52 @@ def find_sublattice_embeddings(
...
+        )
+        if len(defect_free_embs) == 0:
+            # If embedding is infeasible on the tile*, it will be infeasible
+            # on all subgraphs thereof (assuming sufficient time was provided
+            # in embedder_kwargs)
+            return []
+        if use_tile_embedding:
```
thisac:
```suggestion
            return []

        if use_tile_embedding:
```

minorminer/utils/parallel_embeddings.py
jackraymond: Timeouts smaller than 1 are allowed for these functions. A separate timeout can be given as part of embedder_kwargs where there could be problems?

minorminer/utils/parallel_embeddings.py
jackraymond: Should be an empty dictionary (object of length 0) when it fails, added to docstring.

thisac (approved):
LGTM. Thanks @jackraymond!

PR #266, opened 2025-06-04
Parallel embeddings modification to accommodate composite failure to propagate graph type information

jackraymond:
Allow lattice type and lattice dimensions to be passed as options for find_sublattice_embeddings.

The embedder might be used by methods in which the graph family cannot be guaranteed, 

e.g. 
https://github.com/dwavesystems/dwave-system/pull/569
If we apply the ParallelEmbeddingComposite to a composite of the QPU, rather than directly, topology information is deleted. We can still use ParallelEmbeddingComposite as a tiling composite if we allow explicit passing of this information


minorminer/utils/parallel_embeddings.py
```diff
@@ -104,12 +104,12 @@ def embeddings_to_array(embs: list, node_order=None, as_ndarray=False):
 
 
 def array_to_embeddings(
-    embs: list, node_order: Optional[Iterable] = None
+    embs: Iterable, node_order: Optional[Iterable] = None
```
randomir:
```suggestion
    embs: Iterable[list], node_order: Optional[Iterable] = None
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -296,9 +294,31 @@ def _mapped_proposal(emb: dict, f: Callable, one_to_iterable: bool = True):
...
+        elif T_family == "zephyr":
+            return dnx.zephyr_graph(**T_kwargs)
+        else:
+            raise ValueError("Unsupported dwave_networkx graph")
+
+
 def find_sublattice_embeddings(
     S: nx.Graph,
-    T: nx.Graph,
+    T: Optional[nx.Graph],
```
randomir:
```suggestion
    T: Optional[nx.Graph] = None,
```

randomir: Also, can you update type hints below? I can't make suggestions on lines that haven't been modified.
```python
    tile: Optional[nx.Graph] = None,
    sublattice_size: Optional[int] = None,
    max_num_emb: int = 1,
    use_filter: bool = True,
    use_tile_embedding: Optional[bool] = None,
    tile_embedding: Optional[dict] = None,
    seed: Optional[Union[int, np.random.RandomState, np.random.Generator]] = None,
    embedder: Optional[callable] = None,
    embedder_kwargs: Optional[dict] = None,
    one_to_iterable: bool = False,
    shuffle_all_graphs: bool = False,
    shuffle_sublattice_order: bool = False,
    timeout: float = float("Inf"),
```
I.e. [`Optional`](https://docs.python.org/3/library/typing.html#typing.Optional) denotes `None` is a possible value, not that the argument is actually optional - quite confusing, I know.

minorminer/utils/parallel_embeddings.py
```diff
@@ -404,21 +433,27 @@ def find_sublattice_embeddings(
             embedder=embedder,
             embedder_kwargs=embedder_kwargs,
         )
+    if T.graph.get("family", T_family) not in {"chimera", "pegasus", "zephyr"}:
```
randomir: Since `T` is optional, should check for `None` first.

tests/utils/test_parallel_embeddings.py
```diff
@@ -474,3 +476,67 @@ def test_chimera_on_pegasus(self):
...
+                num_emb = 2
+                T_kwargs = {"coordinates": True}
+            if with_Tfamily is False:
+                T_family = None
+                T_kwargs = None  # Ignored in any case.
+            else:
+                # Cast T as standard nx.Graph; make sure propagation is correct:
+                T_kwargs["m"] = T.graph.get("rows")
+                T = nx.from_edgelist(T.edges)
+            if with_Ts is False:
```
randomir: Wrong variable, that's why the bug for `T=None` is never caught with this test.
```suggestion
            if with_T is False:
```
Or, change `with_T` to `with_Ts` in the for loop above.

Overall, though, I'd use `parameterize` over the test case, not a loop.

tests/utils/test_parallel_embeddings.py
```diff
@@ -474,3 +476,67 @@ def test_chimera_on_pegasus(self):
...
         expected_number_of_cells = (m_nice // m_sub) * (n_nice // 3) * 3
         self.assertEqual(len(embeddings), expected_number_of_cells)
+
+    def test_T_family_T_kwargs(self):
+        # Like test basic, but pass family information
+        with_Ts = [True, False]
+        with_Tfamily = [True, False]
+        for topology, with_T, with_Tfamily in product(
+            ["chimera", "pegasus", "zephyr"], with_Ts, with_Tfamily
+        ):
```
randomir: To use `parameterized`, add `from parameterized import parameterized` at the top, and then:
```suggestion
    @parameterized.expand(
        product(["chimera", "pegasus", "zephyr"], [False, True], [False, True])
    )
    def test_T_family_T_kwargs(self, topology, with_T, with_Tfamily):
        # Like test basic, but pass family information
```
Also, need to dedent the ex for body.

tests/utils/test_parallel_embeddings.py
randomir: Added the code suggestion for parameterized use.

minorminer/utils/parallel_embeddings.py
jackraymond: Which argument marked as Optional cannot be None?

jackraymond: I see, you are giving me the corrected format

randomir (approved):
Perfect, thanks!

PR #275, opened 2025-09-16
Add keyboard interrupts for subgraph.find_subgraph

boothby:
I've finally figured out a (hopefully) good-enough means to halt on ctrl-c within `find_subgraph`.  Sadly, testing this capability in CI has proven flaky in the past (as seen in [tests/test_lib.py:841 ](https://github.com/dwavesystems/minorminer/blob/2d5ace1286592449c87c5e7694d8d172fc923c2e/tests/test_lib.py#L841)).  I'll try and scrounge some folks on windows and osx to do some tests.

### DIFF
--- a/minorminer/subgraph.pyx
+++ b/minorminer/subgraph.pyx
@@ -62,6 +62,20 @@ cdef class _labeldict(dict):
         return self._label[k]
 
 
+cdef extern from "<atomic>" namespace "std" nogil:
+    cdef cppclass atomic_bool "std::atomic<bool>":
+        void store(bool)
+
+ctypedef void(*sig_handler)(int) 
+
+cdef extern from "<csignal>" nogil:
+    cdef int SIGINT
+    cdef sig_handler signal(int, sig_handler)
+
+cdef atomic_bool _interrupt_find_subgraph_q
+cdef void _interrupt_find_subgraph(int _):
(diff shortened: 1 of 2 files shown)


boothby:
Test procedure:

1. run the following code, once in a repl (python, ipython, or your favorite IDE) and once as a script (from the commandline: `python test.py` where `test.py` has the code below)
```python
from minorminer import subgraph, find_embedding
from time import sleep
import dwave_networkx as dnx
source = dnx.chimera_graph(8)
target = dnx.chimera_graph(15, coordinates=True)
target.remove_node((7,7,0,0))

emb = subgraph.find_subgraph(source, target)
print("caught an interrupt in find_subgraph")

emb = find_embedding(source, target, chainlength_patience = 1000000, interactive=True, tries=1)
print("caught an interrupt in find_embedding")

emb = subgraph.find_subgraph(source, target)
print("caught an interrupt in find_subgraph")

try:
    sleep(1000)
except KeyboardInterrupt:
    print("caught an interrupt in sleep")
```
2. hit ctrl-c four times, waiting until you see some output before pressing it again
3. the output should look something like
```
^Ccaught an interrupt in find_subgraph
^Ccaught an interrupt in find_embedding
^Ccaught an interrupt in find_subgraph
^Ccaught an interrupt in sleep
```

PR #276, opened 2025-10-16
Add Python 3.14 and remove 3.9

thisac:


tests/test_lib.py
```diff
@@ -25,9 +25,14 @@
...
 import time
 import signal
-import multiprocessing
 import unittest
 
+import multiprocessing
+# Default changed from 'fork' to 'forkserver' for POSIX systems in Python 3.14.
+# See https://docs.python.org/3.14/whatsnew/3.14.html#deprecated for details.
+if os.name == "posix":
+    multiprocessing.set_start_method("fork")
```
thisac, line 34: No the best fix - just relying on the pre 3.14 default - but likely good enough here since it only affects the two tests `test_interactive_interrupt` and `test_headless_interrupt` already marked as skip in CircleCI (weirdly enough they still seemed to run, and fail, in this PR).

thisac: Could also either remove or look into the tests themselves.

randomir: Not sure why the start method makes a difference for these tests, i.e. for handling keyboard interrupt.

Either way, just reverting to the old method _for tests only_ is not enough. We should either move this to the main library code, so forking is always used, or figure out how to fix interrupt handlers for both fork and forkserver methods.

Related, Kelly's working on #275, so should sync these efforts.

randomir (commented):
LGTM, apart from the forkserver changes.

tests/test_lib.py
thisac: Yeah, agreed. This should probably be fixed in the main code as well. This was just as a quick fix since those tests where already marked as skip.

> Related, Kelly's working on https://github.com/dwavesystems/minorminer/pull/275, so should sync these efforts.

👍

randomir:
Release notes missing.

tests/test_lib.py
boothby: Yeah, I couldn't figure out how to get automated testing of keyboard interrupts to work.   It'd be brilliant if you two could sort that out, but afaik we've never had that.

thisac:
Python 3.14 switch to "forkserver" as start-method for multiprocessing (see [here](https://docs.python.org/3.14/whatsnew/3.14.html#deprecated)) requires process to start in new `__main__` (see [here](https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods)). Using old default "fork" for tests to pass. The issue _shouldn't_ affect main code.

PR #285, opened 2026-06-16
Replace `dwave-networkx` with `dwave-graphs`

randomir:
#### Summary

- Use `dwave-graphs` instead of the deprecated `dwave-networkx`.
- Rename `dnx_layout` to `graph_layout`, but keep it as an alias for backwards compatibility.
- Drop NumPy 1.x support.
- Fix examples (python 2 -> 3 updates, missing requirements, missing visualizations)

#### AI AI Generation Disclosure

No AI tools used.


examples/algorithm_demo.py
```diff
@@ -12,20 +12,20 @@
...
-import dwave_networkx as dnx
+import dwave.graphs
 import matplotlib.pyplot as plt
 
 
 # Parameters of the demo
 wait_for_input = False                   # wait for user input to advance to the next step
 G = nx.complete_graph(8)                # source graph
-C = dnx.generators.chimera_graph(2)     # target graph
+C = dwave.graphs.generators.chimera_graph(2)     # target graph
```
thisac: Should be either
```suggestion
C = dwave.graphs.topologies.chimera_graph(2)     # target graph
```
or
```suggestion
C = dwave.graphs.chimera_graph(2)     # target graph
```

minorminer/layout/__init__.py
```diff
@@ -52,7 +52,7 @@ def find_embedding(
             in the 2-tuple applies to S while the second applies to T.
 
             Note:
-                If ``layout`` is a single function and T is a dnx_graph, then the 
+                If ``layout`` is a single function and T is a D-Wave graph, then the
```
thisac: Do we usually refer to graph-like objects as D-Wave graphs? Maybe calling it something like graph-like or even referring to the `GraphLike` type makes more sense.

thisac (approved):
LGTM. I'm assuming we're deliberately leaving `dnx_layout` as is (which I'm fine with)?

Also, three comment referring to `dnx_graphs` in `layout/placement.py`.

minorminer/layout/layout.py
```diff
@@ -237,9 +238,9 @@ def _p_norm_objective(layout_vector, G_distances, dim, p):
 
 
 def dnx_layout(G, dim=None, center=None, scale=None, **kwargs):
```
JoelPasvolsky: Do you want to change all these ``dnx_`` prefixes (about 50) to ``qpu_`` or something?

minorminer/utils/feasibility.py
```diff
@@ -186,31 +186,31 @@ def lattice_size_lower_bound(
...
+            return dwave.graphs.zephyr_graph(m=lattice_size, t=t)
 
         # A lower bound based on number of variables N = (2m+1)*m*4*t
         lattice_size = np.ceil((np.sqrt(2 * N / t + 1) - 1) / 4)
 
     else:
         raise ValueError(
             "source graphs must be a graph constructed by "
-            "dwave_networkx as chimera, pegasus or zephyr type"
+            "dwave-graphs as chimera, pegasus or zephyr type"
```
JoelPasvolsky:
```suggestion
            "dwave-graphs as Chimera, Pegasus or Zephyr type"
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -257,15 +257,18 @@ def find_multiple_embeddings(
...
+    The lattice size is the parameter ``m`` of a ``dwave.graphs`` graph, also
     called number of rows, or in the case of a chimera graph max(m,n). This
     upper bounds the ``sublattice_size`` for ``find_sublattice_embeddings``.
 
     Args:
-        T: The target graph in which to embed. The graph must be of type
-            zephyr, pegasus or chimera and constructed by dwave_networkx.
+        T:
+            The target graph in which to embed. The graph must be of type
+            zephyr, pegasus or chimera and constructed by one of the graph
```
JoelPasvolsky:
```suggestion
            Zephyr, Pegasus or Chimera and constructed by one of the graph
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -355,10 +359,10 @@ def find_sublattice_embeddings(
     Args:
         S: The source graph to embed.
         T: The target graph in which to embed. Either the graph should be of type
-            chimera, pegasus or zephyr and generated by dwave_networkx, or the
-            graph must be cast to one of these types using `T_family` and `T_kwargs`.
-            If `T` is None, then `T` is generated as a defect free graph.
-        tile: A graph constructed by dwave_networkx applied as a displaceable mask
+            chimera, pegasus or zephyr and generated by ``dwave-graphs``, or the
```
JoelPasvolsky:
```suggestion
            Chimera, Pegasus or Zephyr and generated by ``dwave-graphs``, or the
```

minorminer/utils/parallel_embeddings.py
```diff
@@ -355,10 +359,10 @@ def find_sublattice_embeddings(
...
+            graph must be cast to one of these types using ``T_family`` and ``T_kwargs``.
+            If ``T`` is ``None``, then ``T`` is generated as a defect free graph.
```
JoelPasvolsky:
```suggestion
            If ``T`` is ``None``, then ``T`` is generated as a defect-free graph.
```

JoelPasvolsky (commented):
LGTM

examples/requirements.txt
```diff
@@ -1 +1 @@
-dwave_networkx
+dwave-graphs
```
arcondello: I know we didn't have one before, but worth adding a version? Or adding a comment saying we always want to use the latest?

arcondello (approved):

examples/algorithm_demo.py
randomir: Thanks, I missed this in a hurry. Obviously, this is a human-made PR, not AI. :smile:

minorminer/layout/__init__.py
randomir: I used `dwave-graphs` generated graph and similar elsewhere, forgot to update this.

minorminer/layout/layout.py
randomir: No. I'll just alias this function to `graph_layout` or similar. I didn't have a good idea for the alias on first pass.

minorminer/utils/feasibility.py
randomir: These appear in lowercase all over minorminer, so I didn't want to change it in only a couple of places. We can do it in a proper docs pass, for which minorminer is overdue anyway.

examples/fourcolor.py
```diff
@@ -43,15 +43,15 @@
 supplies those additional requirements to `find_embedding`.
 """
 
-from time import clock
+from time import perf_counter
```
thisac, line 46: 👍

minorminer/layout/layout.py
```diff
@@ -326,6 +326,11 @@ def dnx_layout(G, dim=None, center=None, scale=None, **kwargs):
     return layout.layout
 
 
+def dnx_layout(G, dim=None, center=None, scale=None, **kwargs):
+    """Alias for :func:`graph_layout`."""
```
thisac: Should we also deprecate this right away?

minorminer/layout/layout.py
```diff
@@ -326,6 +326,11 @@ def dnx_layout(G, dim=None, center=None, scale=None, **kwargs):
...
+    return graph_layout(G, dim=dim, center=center, scale=scale, **kwargs)
+
+
 def _nx_to_dnx_layout(center, scale):
```
thisac: For the sake of completeness, maybe rename this also?

examples/fourcolor.py
randomir: The example didn't even run because `time.clock` is a Python 2 function. :open_mouth:

thisac: I think I saw `time.clock` still being used in some tests also. :scream_cat:

minorminer/layout/layout.py
randomir: Yes, we can. I figured no point if we're moving layout stuff to dwave-graphs anyway, but no harm in deprecating here.

examples/fourcolor.py
thisac: Here: https://github.com/randomir/minorminer/blob/replace-dnx-with-graphs/tests/test_lib.py#L226-L248

minorminer/layout/layout.py
jackraymond: Why deprecate, why not just keep the same name in the context of the new repo? We need a good name for functions that support Chimera, Pegasus and Zephyr Networkx graphhs more generally, I don't think dnx is a bad choice as an acronym.

examples/parallel_embeddings_demo.py
```diff
@@ -117,18 +120,18 @@ def main():
             value_list
         ), "Duplicate target nodes in embeddings."
 
-        # Visualize embeddings if enabled (waiting for visualization function to be migrated to dwave networkx)
```
randomir, line 120: Updated here because we now have `draw_parallel_embeddings` in dnx/dwave-graphs.

minorminer/layout/layout.py
jackraymond: Or is the decision to replace dnx everywhere by dg or something similar?

examples/fourcolor.py
randomir: Given that `time.clock()` fails with `AttributeError` in Python 3, now I'm more worried why is that code never called.

minorminer/layout/layout.py
randomir: I agree we need a good name for the set of dwave topology graphs, but I'm not sure `dnx` is it. Currently it's overloaded to mean "of `dwave_networkx`, the package".

Also, `nx`/NetworkX shouldn't be in the name, as dwave graphs can be (and are) specified in other formats.

PR #287, opened 2026-07-21
Add O(n log n) LIS via patience sorting

mahdiehmalekian:
Implements longest strictly-increasing subsequence using the classic patience-sort algorithm (piles + binary search + predecessor reconstruction).
**Notes**:

Strictly increasing (uses >= in the binary search, not >) — equal values don't extend a subsequence. This is intentional; relaxing it changes semantics for callers relying on distinct endpoints.
Elements need only support < / >= (tuples work).
No external deps.

AI use: I prompted Claude with my own code based on the psuedocode from Wikipedia <https://en.wikipedia.org/wiki/Longest_increasing_subsequence> which was modified to give strictly increasing output and asked it to clean up the code and document it. I checked the resulting code. I also used Claude to generate the test suite and checked it myself.


boothby (approved):

mahdiehmalekian:
**Note to reviewers:** This is one of several clique-embedder PRs (four up already, including this). Not urgent and not a blocker for the next PR. This one is only depended on a couple of PRs down the road.

minorminer/utils/longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,155 @@
...
+
+The tops of the piles are always sorted left-to-right, which is why we can find
+the right pile with a binary search instead of scanning. To recover the actual
+subsequence (not just its length), each value remembers its predecessor: the top
+of the pile immediately to its left when it was placed. Following predecessor
+links back from the last pile reconstructs the subsequence.
+
+Runs in O(n log n) for the forward pass; reconstruction is O(length).
+
+Implementation notes for reviewers:
```
thisac:
```suggestion
Implementation notes:
```

minorminer/utils/longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,155 @@
...
+__all__ = ["longest_strictly_increasing_subsequence"]
+
+
+def longest_strictly_increasing_subsequence(
+    sequence: list[Any],
+) -> tuple[list[Any], dict[int, int]]:
+    """Find a longest strictly-increasing subsequence of `sequence`.
+
+    Args:
+        sequence: the input values. The elements MUST be mutually comparable --
```
thisac:
```suggestion
        sequence: The input values. The elements MUST be mutually comparable --
```

minorminer/utils/longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,155 @@
...
+            they need to support `<` and `>=` against one another (a total order),
```
thisac:
```suggestion
            they need to support ``<`` and ``>=`` against one another (a total order),
```

minorminer/utils/longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Per pile length, the index (into `sequence`) of that pile's current top --
+    # the smallest top value we've been able to achieve for a subsequence of
+    # that length so far. Slot 0 is a sentinel for "the empty subsequence".
+    smallest_tail_index = [-1]
+
+    pile_count = 0
+
+    for i in range(n):
+        value = sequence[i]
```
thisac: Slightly cleaner.
```suggestion
    for i, value in enumerate(sequence):
```

minorminer/utils/longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,155 @@
...
+    (section "Efficient algorithms").
+"""
+
+from typing import Any
+
+__all__ = ["longest_strictly_increasing_subsequence"]
+
+
+def longest_strictly_increasing_subsequence(
+    sequence: list[Any],
```
thisac, line 80: Not sure about this name. It's fine, but something slightly less generic might be better? Maybe `comparables`? Just `sequence` to me sounds more like a type.

minorminer/utils/longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,155 @@
...
+          * indices: maps each position in `subsequence` to the index it came
+            from in `sequence`, so sequence[indices[i]] == subsequence[i].
+    """
+    n = len(sequence)
+
+    # For each value, the index of the value that precedes it in the growing
+    # subsequence (the top of the pile just to its left when it was placed).
+    # -1 means "nothing precedes it". Following these links back reconstructs
+    # the subsequence. Indexed by position in `sequence` (0..n-1).
+    predecessor = [-1] * n
```
thisac: If switching to enumerate below, you're only using `n` in one place.
```suggestion
    # For each value, the index of the value that precedes it in the growing
    # subsequence (the top of the pile just to its left when it was placed).
    # -1 means "nothing precedes it". Following these links back reconstructs
    # the subsequence. Indexed by position in `sequence` (0..n-1).
    predecessor = [-1] * len(sequence)
```

thisac (commented):
Don't forget to add a releasenotes entry.

minorminer/utils/longest_strictly_increasing_subsequence.py
mahdiehmalekian: The function name already indicates we're returning a **subsequence**, so **sequence** naturally names the input we extract it from. The comparability requirement is documented in the args docstring

minorminer/utils/longest_strictly_increasing_subsequence.py
mahdiehmalekian: Done in [c0b074a](https://github.com/dwavesystems/minorminer/pull/287/commits/c0b074adb5e4543de43a61f8fab67400dfa2f814)

minorminer/utils/longest_strictly_increasing_subsequence.py
mahdiehmalekian: And Longest Increasing Subsequence problem is standard in computer science, see eg. https://en.wikipedia.org/wiki/Longest_increasing_subsequence

thisac: I only meant the `sequence` name, `longest_strictly_increasing_subsequence` is fine. :slightly_smiling_face:

mahdiehmalekian: Great. 
I don't terribly mind changing the name but I think with that function name, then **sequence** is actually the more natural input name.

tests/utils/test_longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,184 @@
...
+        sub, idx = lsis(seq)
+        self.assertEqual(len(sub), _brute_force_length(seq))
+        self._assert_valid_result(seq, sub, idx)
+
+    def test_tuple_elements_lexicographic(self):
+        # tuples are comparable lexicographically
+        seq = [(0, 5), (0, 5), (1, 1), (1, 2), (0, 9)]
+        sub, idx = lsis(seq)
+        self.assertEqual(len(sub), _brute_force_length(seq))
+        self._assert_valid_result(seq, sub, idx)
```
thisac, line 118: It's much preferred to hard-code inputs/outputs instead of relying on a separate implementation of the calculation. Since the expected output `sub` is known, it can easily just be written in and compared to.

I'd probably even test the actual subsequence instead of just the length (since they're deterministic).

tests/utils/test_longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,184 @@
...
+                len(sub), _brute_force_length(seq), msg=f"length mismatch for {seq}"
+            )
+            for a, b in zip(sub, sub[1:]):
+                self.assertLess(a, b, msg=f"not strictly increasing for {seq}")
+            self.assertEqual(set(idx), set(range(len(sub))))
+            src = [idx[i] for i in range(len(sub))]
+            for i, s in enumerate(src):
+                self.assertEqual(seq[s], sub[i], msg=f"index map wrong for {seq}")
+            for a, b in zip(src, src[1:]):
+                self.assertLess(a, b, msg=f"indices not increasing for {seq}")
```
thisac, line 160: Unit tests shouldn't be random (preferably). Better to have a set of different cases that cover all or most edge-cases to make testing repeatable and stable.

tests/utils/test_longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def test_large_input_length_matches_reference(self):
+        # a larger input, checked for length + validity (brute force would be
+        # too slow here, so use a simpler O(n^2) DP as the reference length)
+        rng = random.Random(1)
```
thisac, line 165: Same here with respect to randomness.

mahdiehmalekian: Done in [e2e665d](https://github.com/dwavesystems/minorminer/pull/287/commits/e2e665d68f58cd8f586c50c19112e2e9463c387e)

tests/utils/test_longest_strictly_increasing_subsequence.py
mahdiehmalekian: Done in [e2e665d](https://github.com/dwavesystems/minorminer/pull/287/commits/e2e665d68f58cd8f586c50c19112e2e9463c387e)

tests/utils/test_longest_strictly_increasing_subsequence.py
mahdiehmalekian: Done in [e2e665d](https://github.com/dwavesystems/minorminer/pull/287/commits/e2e665d68f58cd8f586c50c19112e2e9463c387e) and added tests for the actual subsequence and indices. Since this made several other tests redundant, removed those after this addition.

releasenotes/notes/add-lsis-6a27ee13ca010ea7.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Added a ``longest_strictly_increasing_subsequence`` module providing a
+    patience-sort implementation of the longest strictly-increasing subsequence
+    problem in ``O(n log n)``. It returns both one longest subsequence and the
+    mapping from each output position back to its index in the input. Ties are
+    resolved deterministically.
```
thisac: Better refer to the functions that's added.
```suggestion
    Add a ``longest_strictly_increasing_subsequence`` function providing a
    patience-sort implementation of the longest strictly-increasing subsequence
    problem in ``O(n log n)``. It returns both one longest subsequence and the
    mapping from each output position back to its index in the input. Ties are
    resolved deterministically.
```

thisac (approved):
Thanks @mahdiehmalekian !

minorminer/utils/__init__.py
```diff
@@ -20,3 +20,4 @@
 DisconnectedChainError, InvalidNodeError, MissingEdgeError)
 from minorminer.utils.diagnostic import (diagnose_embedding, is_valid_embedding,
 verify_embedding)
+from minorminer.utils._clique.longest_strictly_increasing_subsequence import *
```
randomir:
```suggestion
```
Don't import into public module if we want to keep this function private.

minorminer/utils/_clique/longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,153 @@
...
+        # left-to-right by construction, so a binary search finds the slot.
+        lo = 1
+        hi = pile_count + 1
+        while lo < hi:
+            mid = lo + (hi - lo) // 2  # lo <= mid < hi
+            if sequence[smallest_tail_index[mid]] >= value:
+                hi = mid
+            else:
+                lo = mid + 1
+        pile = lo
```
randomir: This binary search could be replaced with stdlib [`bisect_left`](https://docs.python.org/3/library/bisect.html#bisect.bisect_left), for simplification, but fine to keep.

minorminer/utils/_clique/longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,153 @@
...
+"""
+
+from typing import Any
+
+__all__ = ["longest_strictly_increasing_subsequence"]
+
+
+def longest_strictly_increasing_subsequence(
+    sequence: list[Any],
+) -> tuple[list[Any], dict[int, int]]:
```
randomir: If `sequence[indices[i]] == subsequence[i]` holds, as stated above, then you could/should simplify the return type for indices to be a `list`, not `dict` (cleaner and cheaper).

minorminer/utils/_clique/longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,153 @@
...
+    (section "Efficient algorithms").
+"""
+
+from typing import Any
+
+__all__ = ["longest_strictly_increasing_subsequence"]
+
+
+def longest_strictly_increasing_subsequence(
+    sequence: list[Any],
```
randomir: Could use `Sequence` for type.

tests/utils/clique/test_longest_strictly_increasing_subsequence.py
```diff
@@ -0,0 +1,90 @@
...
+
+class TestLongestStrictlyIncreasingSubsequence(unittest.TestCase):
+    """Behavior and edge cases of the returned (subsequence, indices) pair."""
+
+    # (sequence, expected_subsequence, expected_indices). Every case here has a
+    # UNIQUE longest strictly-increasing subsequence, so the expected output is
+    # forced by the problem -- not by this implementation's tie resolution. That
+    # keeps these as true contract tests: any correct implementation must return
+    # exactly this, and a change to tie-breaking cannot make them fail.
+    @parameterized.expand([
```
randomir, line 31: I don't see a test for duplicate values, like `[3, 3]`, or `[1, 2, 2, 4]`.

Also, no strictly decreasing sequence, like `[3, 2, 1]`.

randomir (commented):
Looks correct, just a few suggestions.

tests/utils/clique/test_longest_strictly_increasing_subsequence.py
mahdiehmalekian: Done in [7cebabd](https://github.com/dwavesystems/minorminer/pull/287/commits/7cebabd3b5ae08142ccfb7c370db0d5e96882d54)

minorminer/utils/_clique/longest_strictly_increasing_subsequence.py
mahdiehmalekian: Done in [7cebabd](https://github.com/dwavesystems/minorminer/pull/287/commits/7cebabd3b5ae08142ccfb7c370db0d5e96882d54)

minorminer/utils/_clique/longest_strictly_increasing_subsequence.py
mahdiehmalekian: Done in [7cebabd](https://github.com/dwavesystems/minorminer/pull/287/commits/7cebabd3b5ae08142ccfb7c370db0d5e96882d54)

minorminer/utils/_clique/longest_strictly_increasing_subsequence.py
mahdiehmalekian: Done in [7cebabd](https://github.com/dwavesystems/minorminer/pull/287/commits/7cebabd3b5ae08142ccfb7c370db0d5e96882d54)

minorminer/utils/__init__.py
mahdiehmalekian: Done in [7cebabd](https://github.com/dwavesystems/minorminer/pull/287/commits/7cebabd3b5ae08142ccfb7c370db0d5e96882d54)

